PR #2, opened 2013-02-07
Ensure div is at top/left when scroll bars removed

davidkethel:
Ensure the panel is at the top/left after an update where the scroll bars
are removed. Other wise it is possible to get stuck midway down a panel
with no way of scrolling.

### DIFF
--- a/src/perfect-scrollbar.js
+++ b/src/perfect-scrollbar.js
@@ -60,6 +60,7 @@
             else {
                 scrollbar_x_width = 0;
                 scrollbar_x_left = 0;
+                $this.scrollLeft(0);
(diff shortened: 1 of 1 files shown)


hatashiro:
Thanks!

By the way, I'd be very appreciated if you change the commit title more specific.
I think "Update src/perfect_scrollbar.js" is somewhat ambiguous.

Regards.

davidkethel:
Cool, no worries

hatashiro:
I think that the pull request's title is changed but the commit message is not changed..
Anyway, I merged it and there's no problem.

Thanks!

PR #3, opened 2013-02-09
 Take into account padding and margins on content

davidkethel:
Changed the height and widths calculations of the content div to
include any padding and/or margins.

There is also some strange behaviour when the container has padding and or margins, which I am still looking into.

### DIFF
--- a/src/perfect-scrollbar.js
+++ b/src/perfect-scrollbar.js
@@ -51,8 +51,8 @@
         var updateBarSizeAndPosition = function() {
             container_width = $this.width();
             container_height = $this.height();
-            content_width = $content.width();
-            content_height = $content.height();
+            content_width = $content.outerWidth(true);
+            content_height = $content.outerHeight(true);
             if(container_width < content_width) {
                 scrollbar_x_width = parseInt(container_width * container_width / content_width);
                 scrollbar_x_left = parseInt($this.scrollLeft() * container_width / content_width);


hatashiro:
Hi!

Thanks for this patch. I'm very appreciated of your patches.

As I tried this, there's still some problem.

In browser, an element's scrollTop property have maximum value, which is outer height(height with padding) + margin-top. This is accepted to scrollLeft too(width with padding + margin-left).

As I tested, in some way with javascript, we can change scrollTop and scrollLeft to larger value manually, but can't do that with mouse wheel(so in our plugin, works well with manually clicking and moving the bar, but doesn't work with mouse wheel).

So when applying this patch, can't scroll to bottom or right fully with mouse wheel.

And even though we can scroll fully with clicking and moving the bar with mouse, the scroll bar's movement looks laggy when the scroll exceed the margin in bottom or right.

I think this problem should be solved before accepting this patch and the solution should be included in the patch.

My object is making this plugin work same with default scroll bar.

So, we should see how the default scroll bar works when the content have margin and padding.

If you find the solution, please let me know with new pull request. I'll look into this too.

Thanks.

Regards.

davidkethel:
Cool, I'll look into the behaviour of the mouse wheel. I have to admit, I made this change on a laptop with only the mouse pad, so i didn't even think of it.

hatashiro:
I fixed a little and pulled this request as commit 35ab2f9.

I changed 'includeMargin' option of outerWidth and outerHeight to false,

because which causes problem is margin, not padding.

So I decided to apply this patch with padding first.

I'll release new version with this and look into margin from now.

Thanks!

PR #12, opened 2013-03-01
Scroll shadows at top/bottom

pateketrueke:
After watching @leaverou talking about this technique using pure CSS, I tried to use this on a project I'm working, but the problem is, actually, that "background-attachment: local" isn't working for all browsers by now.

Issue: http://lea.verou.me/2012/04/background-attachment-local/

I decided to implement this, If you like, let me know.

Sorry for my bad english too.

### DIFF
--- a/src/perfect-scrollbar.css
+++ b/src/perfect-scrollbar.css
@@ -1,7 +1,9 @@
 .ps-container {
-  background-image: radial-gradient(at 50% 0, rgba(0, 0, 0, .1), transparent 70%), radial-gradient(at 50% 100%, rgba(0, 0, 0, .1), transparent 70%);
-  background-size: 100% 10px, 100% 10px;
+  background-image: -webkit-radial-gradient(50%, circle, rgba(0, 0, 0, 0.2), transparent 70%), -webkit-radial-gradient(50% , circle , rgba(0, 0, 0, 0.2), transparent 60%);
+  background-image: -moz-radial-gradient(50%, circle, rgba(0, 0, 0, 0.2), transparent 70%), -moz-radial-gradient(50% , circle , rgba(0, 0, 0, 0.2), transparent 60%);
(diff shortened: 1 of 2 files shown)


hatashiro:
Hi!

First of all, I'm really thank you for your interests about this plugin.
But I'm afraid to say that it's hard to merge this request.

I'd like to make this plugin as small as possible and allow users to customize their own css.
It's the design concept of perfect-scrollbar.
Surely these patches contains some script changes, but I think that the changes are kind of different with my design concept.
I'm afraid but I don't want to add some lines to script for additional design issues.

Rather than take this patches into main repository, I suggest that you can fork this plugin and create your own version with this changes.

Thank you very much again and I'll not close this issue for some days.
If you have some more opinions of this issue, please feel free to let me know.

Regards.

pateketrueke:
You're right!

After made pull-request I realize that my "patch" is a bad feature that breaks your design goal, so I want to rephrase my idea behind:

I just want to add my own css rules if the scroll reach the top, bottom, left or right side.

Well, so the patch should fire an event or just add a css-class?

Thanks for your time.

hatashiro:
I think both ways will cost quite much cpu utilization and it's not what I want with this plugin.

In fact, this plugin uses default 'scrollTop' and 'scrollLeft' properties.
It means, you can write your own events like this, for example:

<pre>
$('#test').perfectScrollbar();
$('#test').scroll(function(e) {
  if( $('#test').scrollTop() == 0 ) {
    $('#test').addClass('top');
  }
  else {
    $('#test').removeClass('top');
  }
});
</pre>

In conclusion, implementing full-featured event trigger or css handling is kind of far from my goal.
I'm very sorry and if you still want the features, please feel free to fork this project and make your own version.

Thank you.

Regards.

---

I updated this comment at 14:11 with test. It works now without any patch!

pateketrueke:
Well done, just perfect.

:+1:

PR #19, opened 2013-03-21
Fixed bug with the Mouse Wheel and the right scroll bar.

EtnasSoft:
When a content has tags inside, like paragraphs, there is a bug with the scroll bar and the mouse wheel. 

When scrolling, the bar doesn't stop at the bottom of the content layer; it continues to fall infinitely. 

With this fix, the behaviour of the right bar is now perfect.

### DIFF
--- a/src/perfect-scrollbar.js
+++ b/src/perfect-scrollbar.js
@@ -90,6 +90,10 @@
                 $this.scrollTop(0);
             }
 
+            if(scrollbar_y_top >= container_height - scrollbar_y_height) {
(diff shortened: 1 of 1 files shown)


hatashiro:
Hi!

Firstly, thank you very much of your patch.

But I can't understand in which situation the bug occurs exactly.

Could you please introduce some example codes?

Regards.

berndschimmer:
works perfect for me! had the same problem after loading dynamic content.

PR #26, opened 2013-05-22
Wrap core functionality in an each function

ahmadsherif:
Quoting the commit message:

> If we have many DOM elements that match the PerfectScrollbar selector, then all elemets get scrolled if one of them is scrolled.

Thanks for this nice plugin by the way.


hatashiro:
Hello.

Thank you for the patch. I'll gratefully merge this and it wil be included in the next version!

Regards.

PR #33, opened 2013-06-11
Fixed some bug in xp sp3 ie7 & ie6 and add an example

maoziliang:
1. fixed mouse scroll over blank area bug.
2. fixed contentHeight and contentWidth retrieve.
3. add an examples based on text content.
4. fixed the unsupported ':hover' pseudo-class  in ie6.
5. fixed the position of scrollbar when scroll in ie6.
6. fixed the size of scrollbar in ie6


maoziliang:
The history is not clear and existing so many commits. So i decide to close this request.

PR #35, opened 2013-06-13
Fixed a few of bugs for ie6~ie10

maoziliang:
#### Fixed list

---
1. Add a sample 
2. Add `ieSupport`. This feature provide the `ie` and `ie<version>` **class** for the `.ps-container` element. It's convenient to figure out the ie's css issues.
3. Add `.hover` for ie6 to support the `:hover` pseudo class.
4. $el.outerHeight() and $el.outerWidth(), also the $el.prop('scrollHeight'), retrieve an incorrect value in ie7 because of the 'hasLayout' problem. 
5. Fixed the follow bug: win7 ie10/xp sp3 ie7 can not scroll the scollbar when the mouse is over the blank area.
6. Fixed scrollbar's height for ie6.
7. Fixed scrollbar's position for ie6.

### DIFF
--- a/src/perfect-scrollbar.js
+++ b/src/perfect-scrollbar.js
@@ -307,9 +307,33 @@
         $this.data('perfect-scrollbar-destroy', null);
       };
 
+      var ieSupport = function (version) {
+        $this.addClass('ie').addClass('ie' + version);
+        if (version === 6) {
+          var mouseenter = function () { 
+            $(this).addClass('hover');
+          };
+          var mouseleave = function () {
+            $(this).removeClass('hover');
+          };
+          $this.bind('mouseenter.perfect-scroll', mouseenter)
+            .bind('mouseleave.perfect-scroll', mouseleave);
+          $scrollbarX.bind('mouseenter.perfect-scroll', mouseenter)
+            .bind('mouseleave.perfect-scroll', mouseleave);
+          $scrollbarY.bind('mouseenter.perfect-scroll', mouseenter)
+            .bind('mouseleave.perfect-scroll', mouseleave);
+        }
+      };
+
       var isMobile = /Android|webOS|iPhone|iPad|iPod|BlackBerry/i.test(navigator.userAgent);
 
       var initialize = function () {
+        var ieMatch = navigator.userAgent.toLowerCase().match(/(msie) ([\w.]+)/);
+        if (ieMatch && ieMatch[1] === 'msie') {
+          //must be executed at first, because 'ieSupport' may addClass to the container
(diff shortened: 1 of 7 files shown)


hatashiro:
I reviewed and commented for each commits.
Please take them into account and apply them to the patch.

Thank you!

src/perfect-scrollbar.js
```diff
@@ -296,12 +296,31 @@
...
+      };
+
       var isMobile = /Android|webOS|iPhone|iPad|iPod|BlackBerry/i.test(navigator.userAgent);
 
       var initialize = function () {
         updateBarSizeAndPosition();
         bindMouseScrollXHandler();
         bindMouseScrollYHandler();
+        var ieMatch = navigator.userAgent.toLowerCase().match(/(msie) ([\w.]+)/);
+        if (ieMatch && ieMatch[1] == 'msie' && ieMatch[2] == '6.0') {
```
hatashiro: Use '===' instead of '=='. '==' is kinda dangerous in JavaScript.

Please refer http://javascript.crockford.com/code.html.

maoziliang:
Ok, I'll correct these issues later.

hatashiro:
I think that it'll be nice to gather the ie fixes together.

How about make a class .ie and .ie6, .ie7, etc?
It'll be added with the script:

``` javascript
var ieSupport = function (version) {
  $this.addClass('ie');
  $this.addClass('ie' + version);
  if (version === 6) {
    // blahblah....
  }
  // blahblah....
};
// ...
var ieMatch = navigator.userAgent.toLowerCase().match(/(msie) ([\w.]+)/);
if (ieMatch && ieMatch[1] === 'msie') {
  ieSupport(parseInt(ieMatch[2], 10));
}
```

If then, we can avoid that the useless CSS styles are applied to the modern browsers.
We can use the css like this:

``` css
.ps-container.ie {
  // some fixes
}
```

Please take this into account.
If it's too time-heavy to implement this feature, just fixes some convention defects and I will implement this feature.

Thank you!
Regards.

maoziliang:
sounds great! weekend is coming. Perhaps I can implement it.

hatashiro:
And I adopted Grunt into this project.

Please refer to https://github.com/noraesae/perfect-scrollbar/wiki/Contributing.

I feel really sorry for the hard review, but as you know, it's necessary for the project maintenance.

Really really thank you very much for your patches!

hatashiro:
I manually merged the patch and linted the codes.
Instead of adding `!important`, I made a patch, 7bd816e79f4cf48f9f7fe56289e1e15e95f03382.

Thank you very much!

PR #36, opened 2013-06-13
use scrollHeight & scrollWidth to avoid getting the wrong size of the content

zenozeng:
use $this[0].scrollHeight & $this[0].scrollWidth to avoid getting the wrong size of the content

### DIFF
--- a/src/perfect-scrollbar.js
+++ b/src/perfect-scrollbar.js
@@ -70,8 +70,8 @@
       var updateBarSizeAndPosition = function () {
         containerWidth = $this.width();
         containerHeight = $this.height();
-        contentWidth = $content.outerWidth(false);
-        contentHeight = $content.outerHeight(false);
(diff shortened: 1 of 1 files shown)


hatashiro:
Hi!

With a quick look, the patch looks nice.

I reviewed the patch and I'd like to make some comments.

First, please remove the lines commented with '//'.
It's the basic refactoring rule that the unused lines should be removed. It looks that they'll be useful someday, but They'll make the code into spaghetti someday.

Second, I think $this.prop('scrollHeight') and $this.prop('scrollWidth') look better with other lines.

Thank you very much for the pull request!
Please take the comments in account and apply them to the patch.

hatashiro:
I'll merge the pull request when I'm home and have some tests with IE.

Thank you!

hatashiro:
Works very well in win7 ie8!

And even, this patch resolve the multiple children problem!!

Thank you very much.

PR #37, opened 2013-06-13
Adding min and max yHeight/xWidth options for scroll bar handle

itsdrewmiller:
This is a feature addition not a bug, so I will not be offended if you don't want to use it.  I found that it was kind of frustrating when navigating across a long list of items to have the thumb of the scrollbar keep shrinking down to only a few pixels.

I updated the readme to include documentation for the options.  I could definitely see adding some defaults for this but I didn't want to break any existing functionality so I did not add them myself.

(Also I had no idea what the part was called - saw it called "thumb" here:  http://my.safaribooksonline.com/book/programming/microsoft-wpf/9780768695519/advanced-scrolling/ch07lev1sec1)

### DIFF
--- a/src/perfect-scrollbar.js
+++ b/src/perfect-scrollbar.js
@@ -69,8 +69,8 @@
       };
 
       var getSettingsAdjustedThumbSize = function (thumbSize) {
-        if (settings.minScrollbarSize) {
-          thumbSize = Math.max(thumbSize, settings.minScrollbarSize);
+        if (settings.minScrollbarLength) {
+          thumbSize = Math.max(thumbSize, settings.minScrollbarLength);
         }
         return thumbSize;
       };
--- a/src/perfect-scrollbar.js
+++ b/src/perfect-scrollbar.js
@@ -7,8 +7,7 @@
   var defaultSettings = {
     wheelSpeed: 10,
     wheelPropagation: false,
-    minThumbSize: null,
-    maxThumbSize: null
+    minScrollbarLength: null,
   };
 
   $.fn.perfectScrollbar = function (suppliedSettings, option) {
@@ -70,14 +69,11 @@
       };
 
       var getSettingsAdjustedThumbSize = function (thumbSize) {
-        if (settings.minThumbSize) {
-          thumbSize = Math.max(thumbSize, settings.minThumbSize);
-        }
-        if (settings.maxThumbSize) {
-          thumbSize = Math.min(thumbSize, settings.maxThumbSize);
+        if (settings.minScrollbarSize) {
+          thumbSize = Math.max(thumbSize, settings.minScrollbarSize);
         }
         return thumbSize;
-      }
+      };
 
       var updateBarSizeAndPosition = function () {
         containerWidth = $this.width();
@@ -188,8 +184,8 @@
 
         $(document).bind('mousemove.perfect-scroll', function (e) {
           if ($scrollbarY.hasClass('in-scrolling')) {
-             updateContentScrollTop();
-             moveBarY(currentTop, e.pageY - currentPageY);
+            updateContentScrollTop();
+            moveBarY(currentTop, e.pageY - currentPageY);
             e.stopPropagation();
             e.preventDefault();
           }
--- a/src/perfect-scrollbar.js
+++ b/src/perfect-scrollbar.js
@@ -6,7 +6,9 @@
   // The default settings for the plugin
   var defaultSettings = {
     wheelSpeed: 10,
-    wheelPropagation: false
+    wheelPropagation: false,
+    minThumbSize: null,
+    maxThumbSize: null
   };
 
   $.fn.perfectScrollbar = function (suppliedSettings, option) {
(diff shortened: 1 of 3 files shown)


itsdrewmiller:
Probably worth noting that I did NOT build the min stuff here - still gotta figure out how to do that.

src/perfect-scrollbar.js
```diff
@@ -67,13 +69,23 @@
...
 
+      var getSettingsAdjustedThumbSize = function (thumbSize) {
+        if (settings.minThumbSize) {
+          thumbSize = Math.max(thumbSize, settings.minThumbSize);
+        }
+        if (settings.maxThumbSize) {
+          thumbSize = Math.min(thumbSize, settings.maxThumbSize);
+        }
+        return thumbSize;
+      }
```
hatashiro: var style function definition should finish with semicolon.

``` javascript
var functionName = function (...) {
  // ...
};
```

Please refer http://javascript.crockford.com/code.html.

src/perfect-scrollbar.js
```diff
@@ -176,8 +188,8 @@
 
         $(document).bind('mousemove.perfect-scroll', function (e) {
           if ($scrollbarY.hasClass('in-scrolling')) {
-            updateContentScrollTop();
-            moveBarY(currentTop, e.pageY - currentPageY);
+             updateContentScrollTop();
```
hatashiro: why are these lines changed?

hatashiro:
It looks very nice.

About the naming, I prefer to use just 'scrollbar' instead of 'thumb' and 'scrollbar' is now used in the codes.
And I think 'size' is kind of ambiguous because typically it means both width and height.
Also, I think 'max' option is not needed.

So, I suggest the option like:

``` javascript
{
  minScrollbarWidth: 100,
  minScrollbarHeight: 100
}
```

If you have some opinions, feel free to comment!

Thank you.

GregDThomas:
> Probably worth noting that I did NOT build the min stuff here - still
> gotta figure out how to do that.

Has anyone given any thought to using node? It makes running tests and
creating min-ified files as simple as running a single command. I'm happy
to do the necessary refactoring if it's considered worthwhile.

Greg

hatashiro:
I once considered to use node to build the scripts but I think it's kinda heavy job for just a jquery plugin.
So I just added build script and jshint file to use.
But I think it's right time to add a way to build and lint the codes.

I'm looking into this: http://gruntjs.com/

I'll do this in this weekend with new version release.

Thank you, Greg.

GregDThomas:
> I'm looking into this: http://gruntjs.com/

Actually, that's exactly what I meant; I've got another plugin at
https://github.com/GregDThomas/jquery-localtime which uses grunt - you may
be able to use that as a starting point.

HTH,

Greg

hatashiro:
@GregDThomas Thanks!

hatashiro:
I adopted Grunt into this project.

Please refer https://github.com/noraesae/perfect-scrollbar/wiki/Contributing.

Thank you!

itsdrewmiller:
I used "size" because it represents width for the X bar and height for the
Y bar.  :-)  but yeah it does seem ambiguous about if it also impacts the
other direction.  Maybe "length"?  With "scrollbar" that implies to me the
whole thing, like somehow the whole bar will only be 100px even if the
container is 400px.  I don't have a strong opinion though, and I can't
imagine anyone actually wanting the latter feature. I agree the max seems
like an option that maybe is a bad feature to have.  How about just:

{
    minScrollbarLength: 100,
}

src/perfect-scrollbar.js
itsdrewmiller: Good call, I do in fact know this but my editor at work doesn't have jslint
or jshint installed :-)

src/perfect-scrollbar.js
itsdrewmiller: I suspect this is my editor autotabbing to 4 instead of 2 - I'll correct
that.

hatashiro:
Sounds good! :D

itsdrewmiller:
Not sure if I am doing this right (this is like my third pull request and first one I am trying to update the code for) but I think the updates are now included in this - I corrected the spacing, missing ';' and changed the name and brought it down to only one setting. I also added an example page.

hatashiro:
1. I can't automatically merge this pull request. Please rebase the branch with updated origin/master.
2. Please make all commits into 2 commits. one is for the minLength patch itself, and the other is for the example.
3. Please remove ',' and also lint the code with Grunt. About Grunt, please check https://github.com/noraesae/perfect-scrollbar/wiki/Contributing.

ghost:
Is this change finished? Scrollbar is very small on large lists.

fantasia949:
+1 for this feature
Hopefully, it'll be available soon :)

itsdrewmiller:
I just deleted my old fork and redid the work on a new one - I still suck
at git but hopefully having just one commit makes things easier to merge.
:-) I followed the contributing guidelines, linted everything and so on so
hopefully there are no issues this time.

itsdrewmiller:
I'm going to delete the old pull request (if I can) - new one is here:

https://github.com/noraesae/perfect-scrollbar/pull/48

PR #40, opened 2013-06-18
Update perfect-scrollbar.js

davidtong:
We should only care about the visible children. Also, doing prop scrollWidth/Height on children prevents including margins into the formula. See test page: http://jsfiddle.net/davidtong/JHYJJ/


davidtong:
The reason of this patch is that if the container has bottom padding, the scrolling area becomes off. There are workarounds to not use this patch so I am closing this pull request for now. If you find the need to revive this patch, let me know...

PR #42, opened 2013-06-28
Using Useragent to test whether touch was enabled was not realiable

zenozeng:
using !!(('ontouchstart' in window) || window.DocumentTouch && document instanceof DocumentTouch); instead

### DIFF
--- a/src/perfect-scrollbar.js
+++ b/src/perfect-scrollbar.js
@@ -352,7 +352,7 @@
         }
       };
 
-      var isMobile = !!(('ontouchstart' in window) || window.DocumentTouch && document instanceof DocumentTouch);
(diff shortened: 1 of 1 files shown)


internalsystemerror:
I agree, this fixes it for me in Google Chrome Version 27.0.1453.116 m on Windows 8 Pro using touch screen on a Lenevo X1 Carbon Touch. I would recommend the strategy of checking for feature support over particular user agents (granted ie is usually a special case). Lovely plugin!

Ping @noraesae

hatashiro:
Looks good.

But why is there `!!`?

Just `(('ontouchstart' in window) || window.DocumentTouch && document instanceof DocumentTouch)` look better to me.

hatashiro:
I agree with the idea that the boolean value should decide whether the device supports touch event.

And how about changing the name of variable into more semantic one?

For example, `supportsTouch` or `isSupportingTouch`.

PR #48, opened 2013-07-17
minScrollbarLength setting

itsdrewmiller:
I couldn't figure out the rebasing stuff so I just deleted my fork and redid the work.  I believe I followed the contributing guidlines correctly as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,10 @@ Default: 10
 If this option is true, when the scroll reach the end of the side, mousewheel event will be propagated to parent element.  
 Default: false
 
-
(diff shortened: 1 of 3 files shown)


ghost:
Scrolling is not good now. Scrollbar reached the end but the list is not at the end. This setting sould also be included in other scrolling calculations.

itsdrewmiller:
Yep - didn't notice this but should be able to fix it.

hatashiro:
I may be able to fix the problem with the suggestion from #50.

Thank you!

PR #52, opened 2013-08-05
memory leak fixes

frozeman:
Hi,

as is used your script in an onepage js app, where dom removal is done not by query but another render engine, i experienced some memory leaks, which can become a problem, when the app runs to long.

this commit should reduce the leaks as global variables get nulled, when perfectScroll gets destroyed.


hatashiro:
Thanks!

PR #63, opened 2013-09-20
Make it AMD compatible

julien:
Make perfect scrollbar work with AMD


hatashiro:
Seems good!

Thanks.

julien:
Thanks you :)

PR #64, opened 2013-09-22
Use vertical scroll for horizontal scrollbar if only horizontal scrollbar is present

jtmalinowski:
It might be slightly more convenient to be able to use vertical scroll wheel for horizontal scrollbar if vertical scrollbar is not present, what do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,11 @@ Default: false
 ### minScrollbarLength
 When set to an integer value, the thumb part of the scrollbar will not shrink below that number of pixels.  
 Default: null
- 
+
+### useBothWheelAxes
+When set to true, and only one (vertical or horizontal) scrollbar is visible then both vertical and horizontal scrolling will affect the scrollbar.
+Default: false
+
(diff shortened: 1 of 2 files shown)


hatashiro:
Sounds good! But how about making it optional? Conventionally, the features that is not supported by native scrollbar have been adopted with optional parameters.

You can find the example of the optional feature here: #48.

And the min files should not be included. Please remove them from the diff.

Thank you!

jtmalinowski:
Done, updated docs too. Let me know if it's alright.

hatashiro:
Seems good!

And for the last one, I'm very sorry but can you add an example code in 'examples' directory?
There're example codes for optional parameters to help people know how to use the option.

Thank you.

hatashiro:
Never mind. I've added an example while refactoring the code :D

Thank you very much!

jtmalinowski:
Was going to take a stab at it today... Anyway, awesome!

PR #73, opened 2013-10-29
Bug for iPad scrolling

EAndreyF:
In iPad 4 with iOS 7 in div with dynamic content in some rare situation with speed object. If currentTime = startTime then speed.x and speed.y become NaN. May be it'll better to check in touchmove bind, but I think we should check variable before use.


hatashiro:
Hello,

Firstly, thank you very much for the contribute.
I think the patches need to be refactored and also I'll test it with my iPad.

Thank you.

azharisubroto:
any update bro?

hatashiro:
718c297ff39475a1ddeb41ef23652abdb1164780 fixes the problem.
Thanks.

PR #77, opened 2013-11-06
Added 4 new setting options (suppressScrollX, suppressScrollY, scrollXMarginOffset and scrollYMarginOffset)

adrianonantua:
Added 4 new setting options:
### suppressScrollX

When set to true, the scroll bar in X axis will not be available, regardless of the content width.  
**Default: false**
### suppressScrollY

When set to true, the scroll bar in Y axis will not be available, regardless of the content height.  
**Default: false**
### scrollXMarginOffset

The number of pixels the content width can surpass the container width without enabling the X axis scroll bar. Allows some "wiggle room" or "offset break", so that X axis scroll bar is not enabled just because of a few pixels.  
**Default: 0**
### scrollYMarginOffset

The number of pixels the content height can surpass the container height without enabling the Y axis scroll bar. Allows some "wiggle room" or "offset break", so that Y axis scroll bar is not enabled just because of a few pixels.  
**Default: 0**

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,22 @@ When set to true, and only one (vertical or horizontal) scrollbar is visible the
 When set to true, the scroll works with arrow keys on the keyboard. The element is scrolled only when the mouse cursor hovers the element.  
 **Default: true**
 
+### suppressScrollX
+When set to true, the scroll bar in X axis will not be available, regardless of the content width
+**Default: false**
+
+### suppressScrollY
+When set to true, the scroll bar in Y axis will not be available, regardless of the content height
+**Default: false**
+
+### scrollXMarginOffset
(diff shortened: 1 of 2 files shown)


hatashiro:
Hi,

I looked into the codes and contents, and it looks really cool.
I'll merge them after I check the features. I'm kinda busy nowadays and maybe it'll take some days.

Thank you very much!

vojkny:
I tried your patch, but it breaks scrolling in certain cases, unfortunately I don't have time to debug it and give you more feedback.

adrianonantua:
I have not come across any problems. Could you assembly a quick fiddle showing the issue?

wooama:
This would be a  really  big improvement.

hatashiro:
I've tested the patch, and it works very well.
Also as I reviewed the codes, there's no problem at first glance.

Thank you very much!

PR #90, opened 2014-01-14
Perfect Scrollbar Issues 89

WebStew:
Hey,

This resolved the problem for me outlined in issue 89. However I have not had time to properly test this patch nor enough time here to be confident it is the best approach. 

Regards
Stu

### DIFF
--- a/src/perfect-scrollbar.js
+++ b/src/perfect-scrollbar.js
@@ -3,559 +3,627 @@
  */
 'use strict';
 (function (factory) {
-  if (typeof define === 'function' && define.amd) {
-    // AMD. Register as an anonymous module.
-    define(['jquery'], factory);
-  } else {
-    // Browser globals
-    factory(jQuery);
-  }
+    if (typeof define === 'function' && define.amd) {
+        // AMD. Register as an anonymous module.
+        define(['jquery'], factory);
+    } else {
+        // Browser globals
+        factory(jQuery);
+    }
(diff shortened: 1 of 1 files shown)


hatashiro:
Hello,

Each commit's diff is kinda wrong cause of tabsize problem. If the second commit is for fixing it, I think it should be amended to the first one. But nevermind, I can amend it when merging them.

But now I'm not sure it's okay to use `$.now()` to identify the plugin object. It'll be very nice if there's any api that returns the unique identifier for a specific jQuery object, but I don't have any idea yet.

Anyway, the patch looks very good. I'll merge it after I have a little more time to think about the patch.

Thanks!

WebStew:
Hey,

Yeah my apologies on the bad commit. My IDE screwed your formatting. I know what you mean about the UID approach and I'm not sure it is the best way to resolve the issue however I had a deadline here and had to patch it as quickly as possible. I will endeavour to come back to this and see if I can find a better approach. If I do, I'll update the src and recommit on this fork.

Regards
Stu

hatashiro:
Hello,

The issue is resolved with the patch 771ed0c32f0d30b228f4aad4f68432ce2ac1ecc4.
It implements the function to return a unique event class name.

But your report and patch really helps me to do the job. Thank you very much.

WebStew:
Hey,

No problem at all. Glad it helped :)

PR #93, opened 2014-01-21
Hide scrollbar when the container size is equal or smaller to the content's

murillobit:
Fix #71

### DIFF
--- a/src/perfect-scrollbar.js
+++ b/src/perfect-scrollbar.js
@@ -161,6 +161,8 @@
           scrollbarXWidth = 0;
           scrollbarXLeft = 0;
           $this.scrollLeft(0);
+          
+          settings.suppressScrollX = true;
         }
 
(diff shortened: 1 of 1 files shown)


hatashiro:
Hello,
Thank you for the pull request, I'm very sorry but we shouldn't merge this.

First, there's no reset code to set the setting value to false again when the content size is changed and becomes bigger than the container size. Also, I think that changing the settings doesn't make sense. The setting values should be fixed to the values a user set.

But I just created a patch inspired by your patch: db6c0bd968a5e3d3eca8113cc74d0ccc45f989e3.
This patch fixes the problem.

Thank you again for your help :D

murillobit:
Hello,

No problem :)
Actually, I didn't realize about that problems with the settings values. Anyway, it was the first thing that came up to my mind and solved my problem (since my container size never changes).

Thanks for the fix, I'll update my code.

PR #99, opened 2014-01-28
Fixed typo in CSS: missing space in "-webkit-transition:background-color.2s linear".

Modified:
Noticed this because my Chromium complains, and (sometimes?) seems to choke on this bug.


hatashiro:
Thanks!

PR #104, opened 2014-02-10
Don't scroll by keyevent while form element has focus

rangermeier:
Prevent scrolling when keypress event happens inside a form element (input, textarea). Currently perfectScrollbar scrolls the page when a space is entered in an input and prevents the event. This small fix solves this problem.


bkonetzny:
Code works, thank you!!! :)
As a temporary workaround, setting "useKeyboard=false" will also allow correct input.

hatashiro:
Looks good. Thanks!

PR #110, opened 2014-02-19
Add event triggers when scroll element reaches a bounding point

joshuaballoch:
The use case:
1. You want to know when a perfect-scrollbar element has reached the top/bottom/left/right-hand most position it can scroll.

Why:
1. In case other code needs to execute when that bounding point has been reached

Further information:
1. My use case is: I want to use the perfect-scrollbar in an element that will fetch more records when it reaches the bottom of the scroll element.

I have added an example of some simple event triggers that will trigger on the original DOM element, as a suggestion on how to satisfy this use case. 

I have not altered any other code, I want to see what the consensus is on this kind of event trigger first.


hatashiro:
As you can find in `examples/reach-the-end-callback.html`, in fact without an additional event, we can implement this. But I've got quite many requests for this feature, so someday I'll consider it.

Thanks.

PR #111, opened 2014-02-19
Overflow:hidden is a required css style

joshuaballoch:
In testing this plugin, I noticed that another requirement is that the container div MUST have the overflow:hidden style.

I added this to the README, I think it will be useful for others to know this.


betesh:
Thank you, that was helpful!

hatashiro:
Hello,

Thank you for the patch, but the patch removes trailing spaces.
In Markdown format, 2 trailing spaces mean a line break, and shouldn't be removed.
When you fix them, I'll merge the patch.

joshuaballoch:
@noraesae updated. thanks for the awesome plugin, it really is great!

PR #115, opened 2014-02-28
Fix delta calculation in mouswheel handler

rangermeier:
Remove division by 10 of deltas, which makes scrolling slower than in previous versions.
Reverts 88d286957766e981a4f4c720f213164fc0f3f099

Tested with IE11, Firefox 27 and Google Chrome 33 - 35.

I don't know why this division was introduced, also couldn't find a hint on the jquery-mousewheel changelog that this is necessary. Also see examples/options-wheelSpeed.html - 'Slow: wheelSpeed:1' doesn't scroll at all if the delta values are divided by 10.

### DIFF
--- a/src/perfect-scrollbar.js
+++ b/src/perfect-scrollbar.js
@@ -273,8 +273,8 @@
       var bindMouseWheelHandler = function () {
         var shouldPrevent = false;
         $this.bind('mousewheel' + eventClassName, function (e, deprecatedDelta, deprecatedDeltaX, deprecatedDeltaY) {
-          var deltaX = e.deltaX ? e.deltaX / 10 : deprecatedDeltaX,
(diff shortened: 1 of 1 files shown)


kariem:
Thank you!

Just wanted to file the same pull request, although I wasn't sure what to say about that (now reverted) commit that cost me about two hours. Could not find any indication on the reason of that division by 10.

hatashiro:
Hello, the reason for the previous patch is that the scrolling is abnormally fast without the delta divided with 10 in OS X Chrome. I think there're something wrong with Macbook Trackpad in jquery-mousewheel. I didn't notice that the change makes the scrolling slow in other environments. I'm sorry.

I consider to make the plugin independent of jquery-mousewheel, but don't have much time to work it. I'll try to do it asap.

Thank you.

PR #119, opened 2014-03-10
Change mousewheel delta calculation logic.

hatashiro:
Hello,

Recently, we moved to jquery-mousewheel 3.1.9, and it changes the way calculating the deltas. So I changed the logic a little before, but it made the scrolling extremely slow in some environments. So the change was reverted.

I just found that there's a value called `event.deltaFactor` in jquery-mousewheel, and I re-change the logic using `event.deltaFactor`. To prevent the problem like before, I hope to test it in many environments before applying it into the master branch. The change is tested with my Macbook Trackpad and Logitech Mouse in the latest Safari and Chrome. If there's no problem in other environments and seems good, I'll merge the patch.

You can test it with the example, `examples/options-wheelSpeed.html`. The terminal commands will be like this:

```
git clone https://github.com/noraesae/perfect-scrollbar.git -b mousewheel-speed
cd perfect-scrollbar
open examples/options-wheelSpeed.html
```

If there's any problem, please report it with comments. Any further contribution also will be welcomed.

Thanks in advance.

### DIFF
--- a/src/perfect-scrollbar.js
+++ b/src/perfect-scrollbar.js
@@ -277,10 +277,16 @@
 
       // bind handlers
       var bindMouseWheelHandler = function () {
+        // FIXME: Backward compatibility.
+        // After e.deltaFactor applied, wheelSpeed should have smaller value.
+        // Currently, there's no way to change the settings after the scrollbar initialized.
+        // But if the way is implemented in the future, wheelSpeed should be reset.
+        settings.wheelSpeed /= 10;
+
         var shouldPrevent = false;
         $this.bind('mousewheel' + eventClassName, function (e, deprecatedDelta, deprecatedDeltaX, deprecatedDeltaY) {
-          var deltaX = e.deltaX || deprecatedDeltaX,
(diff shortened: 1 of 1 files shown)


le717:
I tested this on my computer (laptop) the other day (Windows 8.1, Firefox 27) using both my mouse and touchpad. There did not _appear_ to be any issues. From what I saw in the ~5 minutes I play with it, everything scrolled at the proper speed both horizontally and vertically. Can you provide some sort of quick visual of the expected speed?

hatashiro:
I changed some contents.

First, while looking the code, I felt a little weird for `wheelSpeed` applied to the keyboard scrolling. I changed keyboard scrolling not to use the `wheelSpeed`.

Second, if the default `wheelSpeed` is changed, backward compatibility problem can occur for the users who upgrade to the upcoming version. So I reverted the patch and decide not to change the default `wheelSpeed`.

The patches will be applied soon with some tests.

Thanks.

hatashiro:
Merged.

PR #131, opened 2014-04-01
Use innerWidth/Height instead of outerWidth/Height to determine the width/height of container

Feasul:
As per jQuery docs the `innerWidth` method is the one that includes padding (https://api.jquery.com/innerWidth/), while the previously used `outerWidth` includes border width as well (https://api.jquery.com/outerWidth/). Therefore I think the use of `innerWidth` makes more sense for getting the container width with padding in the code (plus a setting name `includePadding` suggest only padding width is taken into account as an extra for container width determination and might be confusing otherwise).

Thanks, please excuse me if I was not attentive and missed something from the code.

### DIFF
--- a/src/perfect-scrollbar.js
+++ b/src/perfect-scrollbar.js
@@ -152,8 +152,8 @@
       };
 
       var updateBarSizeAndPosition = function () {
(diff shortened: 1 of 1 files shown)


hatashiro:
Correct. Thanks!

PR #132, opened 2014-04-04
Prevent scrollbar rail disappearance during scrolling

Feasul:
This update to styles makes scrollbar rail stay visible while you drag the scrollbar along it and accidentally leave the area of the rail with mouse pointer (this should not affect scrolling with mousewheel)


hatashiro:
Yes, this patch should be applied.
Thank you!

PR #139, opened 2014-04-16
Added alwaysVisibleX and alwaysVisibleY options

asiviero:
Hello @noraesae, we've been using perfect scrollbar in our website but we needed it to be always visible in some blocks, so I added a couple of options to make it possible.

Hope you like it :D

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,14 @@ The number of pixels the content height can surpass the container height without
 When set to true, it uses `innerWidth` and `innerHeight` for the container size instead of `width` and `height`. When your container element has non-zero padding and the scrollbar layout looks weird, this option can be helpful.  
 **Default: false**
 
+### alwaysVisibleX
+When set to true, the scroll bar in X axis will be visible as if mouse is over the block in which the scroll bar is.
+**Default: false**
+
+### alwaysVisibleY
+When set to true, the scroll bar in Y axis will be visible as if mouse is over the block in which the scroll bar is.
+**Default: false**
+
 How to Use
 ----------
 
--- a/src/perfect-scrollbar.js
+++ b/src/perfect-scrollbar.js
@@ -574,12 +574,12 @@
 
       var supportsTouch = (('ontouchstart' in window) || window.DocumentTouch && document instanceof window.DocumentTouch);
       
-      var setAlwaysVisibleX = function() {
-    	$this.find('.ps-scrollbar-x-rail').css('opacity',0.6);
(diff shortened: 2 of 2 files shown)


hatashiro:
You can do this by changing the css without any script change.
It's a bit changed now but you can refer to the issue #27 

Thanks.

asiviero:
Wouldn't it be simpler if everything could be set in a single place for the scrollbar?

hatashiro:
I'm very sorry but I don't want to have something about CSS style in the script. I think that they should be separated however the implementation is very simple. It's not about functionality, but about design pattern.

I'm sorry again and thank you very much for your interest in my plugin.

asiviero:
No problem at all, the plugin is really great :D

hatashiro:
Thank you for your kindness :)

juanghurtado:
Is this still the way to do it? My `ps-scrollbar-x-rail` has a `display: none` (and `opacity: 1`), and changing it to `block` on my CSS doesn't seems to work because the "drag bar" is not shown. The maths for the width of the "drag bar" seems to begin after first scroll on the scroller, so it can't be visible on page load.

Am I doing something wrong?

mantia11:
@juanghurtado the same

hatashiro:
Hi,

Sorry for the incorrect information. Actually it's nothing to do with the `display` css. Instead, please change every reference of `opacity: 0` to `opacity: 0.6`.

Also, in the development version, The stylesheet source is changed to use scss. There's an instruction about the problem for the new source in [FAQ](https://github.com/noraesae/perfect-scrollbar/wiki/FAQ).

Thanks.

mantia11:
Thanks, now it works okay!

kpturner:
What if you want the scrollbar to be permanently visible in some containers and not in others.  Changing the CSS does not really lend itself to that sort of requirement does it?   It would be easier, and more intuitive, for this to be an option on the plug-in wouldn't it?

hatashiro:
@kpturner If two scrollbars should be shown differently, I think it'd be better to have different CSS styles for each container. I'll update the FAQ and add information to achieve this more easily.

Thanks!

hatashiro:
@kpturner I've just updated [FAQ](https://github.com/noraesae/perfect-scrollbar/wiki/FAQ) and add some information. I also added an example code, `always-visible.html`, in a `examples` directory to help users understand how it works. I hope it could help you.

Thanks.

PR #159, opened 2014-06-19
Add ps-visible class on container then scrollbar is active

sarunas:
Add ps-visible class on container then scrollbar is active


nikaspran:
:+1:

hatashiro:
Looks reasonable.
Before I merge this patch, could you make the patch into just one commit, and exclude the `min` files in your commit?

Thanks.

sarunas:
Updated PR

PR #168, opened 2014-07-01
keydown events for ctrl+home and ctrl+end, ctrl+mousewheel zoom when "mid-scroll"

wikenator:
To allow keyboard scrolling to the extreme top and bottom of a long page, respectively.


wikenator:
Added code to allow mousewheel zoom with ctrl+mousewheel by destroying then recreating perfect scrollbar on affected area of zoom.  Zoom was not possible "mid-scroll" before; zooming would propagate after scrollbar reached the top-most or bottom-most of the container.  Not sure if there is an easier way to detect a control key press to 'ignore' scrolling without having to destroy and recreate the scrollbar..

hatashiro:
I'd like to accept the `ctrl+home` and `ctrl+end` event first, because it makes sense. And about ctrl+mousewheel, I think there's other way instead of destroying the plugin.

Could you make this PR only for the `ctrl+key` feature with just ONE commit, amending a lint commit, and upload a separated PR for zooming?

Thanks.

hatashiro:
I couldn't find your original repo, so I rewrote it by myself.
The patch's been applied, 1b1195288bab3cd49356e36568bbc00acc279180.

Thanks.

PR #177, opened 2014-07-11
Fixes #176

kty:


hatashiro:
Merged manually to add a commit message and squash them into one commit.

Thanks!

PR #180, opened 2014-07-15
Keydown now checks if the event was preventDefault()'ed

thany:


hatashiro:
Merged manually because there was a trailing space.

Thanks!

PR #181, opened 2014-07-15
Apply .ps-active class when area scrollable

andrejcremoznik:
For a project I'm currently working on I needed to know when the area perfect scrollbar is applied to is actually scrollable so I can adjust the styling for the containing elements e.g. I want to remove the padding that the scrollbar covers when it's not active.

This piece of code works fine for me but I admit it's a quick fix. There might be a better place to put it. I do think it's a useful feature though and should be a part of this plugin.


hatashiro:
Duplicated with #159. Anyway, thank you very much for your support!
You can find the feature in the latest version soon.

andrejcremoznik:
Great minds think alike :) Glad to see this going in

PR #200, opened 2014-08-26
IE8 fix - reset railBorderXWidth and railBorderYWidth

mciastek:
In IE8 railBorderXWidth and railBorderYWidth are NaNs. If you reset them (set theirs value to 0) everything is working perfectly.


hatashiro:
Now just hide rails if they're suppressed.

Thanks anyway.

PR #205, opened 2014-09-10
Compatible with iframe

mnico:
Hi, I used this wonderful plugin in my project. The only downside was using iframes. In my project I created a iframe dynamically with many elements and I wanted to attach the plugin inside the iframe. It worked but not the drag and drop feature for the bar. So I made a little adjustment to make it work and i added an example. I hope this serves.


hatashiro:
Manually merged with 6ac02fae2be743d5ab4744d12b4b386e8af3d44e and 467cdb17ed8015e3206a95816fa09506abcf9fb7.

Thanks!

PR #207, opened 2014-09-11
Remove the illogical requirement that overflow:hidden be used on scrollable areas

andrewluetgers:
This makes the scrollbar have fewer requirements and be much more of a drop in solution as you don't need to go around setting all your scroll areas to overflow hidden when you first add it to your app. This also makes more sense in light of the new option (my other PR) to disableIfOSBarsOverlay as in iOS or common configurations of Mac OS which are often preferable to JS sliders.


andrewluetgers:
closed as it made more sense to put this in with my other PR

PR #214, opened 2014-09-23
Added gh page cdn information

genediazjr:
mirror on [cdnjs.com](http://cdnjs.com/libraries/jquery.perfect-scrollbar)


hatashiro:
I think CDN info in GH pages is currently not needed, as we've added it in README.
If I include `install` section in GH pages, I will move all of its options together from README, maybe in the future.

Thanks anyway!

genediazjr:
yup. no prob.

PR #217, opened 2014-09-29
Minor corrections in example.

josemalonsom:
Just a couple of minor corrections in the CSS of the example "How to use".


hatashiro:
Thanks a lot!

Issue #103, opened 2014-02-09
Perfect Scrollbar not working properly inside twitter bootstrap tabs


jagjot2008:
I observed that when i place the code under twitter bootstrap tabs the scrollbars where not showing on hover. However when i hit the down page arrow key they appear.

noraesae:
Closed for the inactivity.

PR #219, opened 2014-10-01
ShadowDom form elements are considered for preventing scrolling by keyevent

MathieuLorber:
I recently had a problem with form elements and scrolling with keys - I couldn't have spaces in my textareas as perfect-scrollbar handled the space keydown event. I posted a message in #103 , been redirected to #123 . My problem should not have occurred...

I finally realized jquery could not identify my textarea as an input element because I use shadowdom and document.activeElement return the shadowdom element and NOT the "final" element. See : http://www.w3.org/TR/shadow-dom/#active-element .

This PR resolves the problem. I thought it deserved a comment, I made it simple as the other comments in the code are mostly short.

I'm not sure this is the only problem perfect-scrollbar could have with shadowdom. I'll come back if I discover any other one...

### DIFF
--- a/src/perfect-scrollbar.js
+++ b/src/perfect-scrollbar.js
@@ -431,7 +431,16 @@
             return;
           }
 
-          if (!hovered || $(document.activeElement).is(":input,[contenteditable]")) {
+          if (!hovered) {
+            return;
+          }
+
+          var activeElement = document.activeElement;
+          // go deeper if element is a webcomponent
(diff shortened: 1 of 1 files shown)


hatashiro:
I really appreciate your contribution! It looks perfect. I'll test it out and merge it in a while.

Cheers

hatashiro:
Merged. Thanks!

MathieuLorber:
Thank you =)

PR #222, opened 2014-10-13
IE Pointer Events and Classes

loliver:
Added support for PointerEvents (and MSPointerEvents) to support touch in IE10 and 11. This is in response to bug #146.

Also added classes to the container when at the limits of the scrollable area, mainly for the implementation we're using, but can be used to show and hide scroll masks and the like.

Apologies for the indentation issues, looks like Visual Studio auto formatted everything.

### DIFF
--- a/src/perfect-scrollbar.js
+++ b/src/perfect-scrollbar.js
@@ -2,717 +2,873 @@
  * Licensed under the MIT License
  */
 (function (factory) {
-  'use strict';
+    'use strict';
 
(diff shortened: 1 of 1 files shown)


hatashiro:
The patch's been manually merged, e6590da884e246046a78edcabf60c3d200ef5e46.
I found that the patch reuse the code in mobile touch handler, so I just implemented the feature in it.
Despite modification by me, I really appreciate you for your idea.

Thanks

PR #225, opened 2014-10-19
Inner perfect-scrollbar fix

ronihcohen:
As you can see at the example "inner.html" I have perfect-scrollbar inside perfect-scrollbar,
When the mouse go over the parent div the children perfect-scrollbar appears.
The problem is that CSS :hover selector is being used and it is referring to all perfect-scrollbar classes in the page.

I fixed that problem by adding "hover" class to $this , and removing the :hover selectors.


hatashiro:
Hi,

Thank you very much for the PR. However, I'd just go with `>` CSS selector. The problem's been just resolved with 65869821c1f14febf42e935aecaa3e59e16b0819.

Cheers,

PR #239, opened 2014-10-30
Don't handle mouse pointer events as touch in IE10

darthmaim:
IE10 returns `MSPOINTER_TYPE_MOUSE` instead of `"mouse"` as source for pointer events (http://msdn.microsoft.com/en-us/library/ie/hh772359.aspx).

In case `event.MSPOINTER_TYPE_MOUSE` is `undefined` (all browsers but IE10) the added comparison will always resolve to true, because we already checked that `event.pointerType` is not undefined earlier.


hatashiro:
Thanks.

PR #251, opened 2014-11-14
Renamed 'int' helper in 'getInt' #250

tatarinov:
Fix error compiling in 'closure' compiler


hatashiro:
I didn't notice `int` in JavaScript can cause a problem. Thanks.

PR #252, opened 2014-11-17
Detects margins around the rails

tkhyn:
Hi and thanks a lot for this perfect scrollbar!

I had the need to add margins at the top and bottom of the Y rail, so I made the changes to enable this functionality after unsuccessfully trying to find a workaround with borders.

The only thing needed is to add `margin-*` CSS properties to the `ps-scrollbar-*-rail`.

I'd say it could be a good thing to have that functionality in the next release to make this plugin even more perfect, but that's for you to say.


hatashiro:
Works like a charm.

Thank you very much!

romandrahan:
@tkhyn, not working with latest perfect-scrollbar version. Adding `margin-top` or `margin-bottom` to the `ps__scrollbar-y-rail` makes scrollbar not working.

@noraesae, I think this feature was removed during upcoming releases?

PR #254, opened 2014-11-19
Add disablePropagation option.

MasGaNo:
I added new option for perfect-scrollbar to disable prevent propagation during scrollbar update. By default, this options is set to true to keep compatibility.
I need to enable propagation even if the scroll doesn't reach the end of the side (with wheelPropagation option), to be able to listen this event on my code to trigger some action.
I hesitated between trigger some built-in event, but I didn't want to introduce some new logical feature in your work, so I extended it by adding this option.
With this option, the propagation will be controlled in my-side, after triggering my actions.

### DIFF
--- a/src/perfect-scrollbar.js
+++ b/src/perfect-scrollbar.js
@@ -28,6 +28,7 @@
   var defaultSettings = {
     wheelSpeed: 1,
     wheelPropagation: false,
+    disablePropagation: true,
     minScrollbarLength: null,
     maxScrollbarLength: null,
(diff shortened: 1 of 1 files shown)


hatashiro:
The plugin structure's changed quite a lot, so the patch can't be merged. I also think that there's already a `wheelPropagation` option, and it's too much to add another option to disable the propagation, even though it can cover additional cases.

Sorry I couldn't take care nowadays and thank you very much for your PR.

Cheers,

PR #270, opened 2014-12-16
Define the used colors as variables

DanielApt:
To enable easy customisation it's nice to have all the used colors as seperate variables at the top of the file.


hatashiro:
Fair enough. Thanks!

PR #283, opened 2015-02-09
No longer remove classes at the beginning of updateGeometry()

michalc:
Chrome and Firefox can be very slow when finding heights and widths
for large scrollable divs just after adding or removing classes.
Since the scrollbars should not affect computed height and width,
there is no need to remove the classes that makes them visible at
the beginning of updateGeometry().

### DIFF
--- a/src/perfect-scrollbar.js
+++ b/src/perfect-scrollbar.js
@@ -73,7 +73,7 @@
       // Catch options
       if (option === 'update') {
         if ($this.data('perfect-scrollbar-update')) {
-          $this.data('perfect-scrollbar-update')();
+          $this.data('perfect-scrollbar-update')(true);
         }
         return $this;
       }
@@ -204,7 +204,12 @@
         $scrollbarY.css({top: scrollbarYTop, height: scrollbarYHeight - railBorderYWidth});
       }
 
-      function updateGeometry() {
+      function updateGeometry(hideRails) {
+        if (hideRails) {
+          $scrollbarXRail.hide();
+          $scrollbarYRail.hide();
+        }
+
         containerWidth = settings.includePadding ? $this.innerWidth() : $this.width();
         containerHeight = settings.includePadding ? $this.innerHeight() : $this.height();
         contentWidth = $this.prop('scrollWidth');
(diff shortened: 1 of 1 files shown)


hatashiro:
Actually they do affect in some special cases. Please refer to #201. Specifically, although they are absolute-positioned, they unfortunately affect scroll width and height and it makes the calculated values wrong in `updateGeometry`. I hide the scrollbar in the beginning of the method by that reason, not to affect the calculation. However, as you said, it may harm the performance hard, so I will find another way to cope with it. I'll leave this PR to track the work.

Thank you very much!

michalc:
@noraesae So I can think about it as well, what are the special cases?

hatashiro:
Hmm, it's kinda difficult to explain, though it's described in the issue #201.

The #201 issue was that when we scroll down til the bottom, increase a container's height and call `.perfectScrollbar('update')`, the scroll exceeds the bottom of the actual content and a white area is shown below because scrollbars remain in the same position as they were before updated and it makes `scrollHeight` larger than the content's actual height. So I just decided to hide(`display: none`) the scrollbars before updating, calculate with correct `scrollWidth` and `scrollHeight`, and show again.

I'll quickly made a fiddle to explain how an absolute-positioned element affect a `scrollHeight` property. Give me a second :)

hatashiro:
I've just written the exact case example! Please refer to the link below.

http://jsfiddle.net/q1fccyzz/2/

Please scroll the containers to their bottom and click the reduce buttons. It clearly shows that an absolute-positioned element does affect the `scrollHeight`. I've also added a button which hide and show the absolute-positioned element before and after updating, and it works fine.

Well, that's why I added the lines even though it harms performance.

Thank you very much for your help :D

michalc:
I think I understand. So how about only hiding the bars on a call to `update`, like in my most recent commit?

hatashiro:
I think it really makes sense! I will apply it in the `refactoring` branch, as it will be merged into the master branch as the next version.

hatashiro:
I pushed the patch into the new master branch with b4a51fc. Thanks for the idea.

PR #286, opened 2015-02-18
False selection detection when typing something into input

rangermeier:
`selection.rangeCount` will return 1 when user is typing something in a
form widget. This will consequently trigger unvoluntary scrolling when
the mouse is moved outside of scrollable area. Checking the actual
length of selected text doesn't trigger this behaviour.

The undesired behaviour can be seen here: http://jsfiddle.net/o2pj5ppz/2/embedded/result/
Same setup with the changes from this commit: http://jsfiddle.net/o2pj5ppz/4/embedded/result/


darthmaim:
:+1:

hatashiro:
Fare enough. Thanks!

PR #289, opened 2015-02-25
Disable selection handler after keyup

michalc:
Fixes a bug where after after entering a character in a input, mouse movement caused scrolling as through the user was selecting a range of text.


michalc:
I've now realised that in the latest version of `perfect-scrollbar`, this pull request isn't required, as the `getRangeNode` function returns null after entering a single character in an input, so the plugin doesn't think any text is selected, and so subsequent mouse movement doesn't cause the bug described. Closing :-)

hatashiro:
Oh, thanks anyway!

Issue #277, opened 2015-01-31
Help needed for the next version!


noraesae:
As you can find in the current master branch, the next version of the plugin will be independent from jQuery and NPM and Browserify friendly! I've done some work on it and it's about to finish, but I'd like somebody to help and participate in the project! Every kind of contribution would be really appreciated.

Here is a TODO list.

- [x] Migrate into Gulp.
- [x] Use Browserify to make the source tree modular.
- [x] Move to eslint.
- [x] Be independent from jQuery.
- [x] Rewrite examples with vanilla JS.
- [x] npm with browserify compatibility.
- [x] Make handlers modular.
- [x] Other adaptors including AMD.
- [x] Include built files in the NPM registry.
- [x] Automatic release with built files into GitHub Releases.
- [x] Separate a bower repo.

Specifically,

1. While modularising it into several part, I detached AMD support temporarily. To be honest, I'm not familiar enough to AMD or other adaptors. Currently, jQuery and just `global` adaptors have been implemented. Any contribution on new adaptors will be gratefully welcomed.
1. I've changed code base quite much, so I'd like to check if it works fine in most environments including IE>=9. I've just checked with my environment(the latest versions of Safari, Chrome and Firefox in OSX) and it looks working fine. Nevertheless, I'd like to make it sure. The easiest ways to test it would be with NPM or git.

NPM
```
$ npm install git+ssh://git@github.com:noraesae/perfect-scrollbar.git
```

git
```
$ git clone git@github.com:noraesae/perfect-scrollbar.git
```

Thanks in advance!

Nijikokun:
Why separate the bower repo, this makes no sense?

noraesae:
Why not? React has [a separated repo](https://github.com/reactjs/react-bower) for Bower. jQuery [has also dropped Bower](https://github.com/jquery/jquery/commit/e1949f43711b5808d42378bafb6728b40b1738d6). Frontend devs are moving to using npm to manage their dependencies and packages, with some help of modules like Browserify or Webpack, and I totally agree to the movement.

I think Bower is being deprecated. However, I don't want to frustrate users who are currently using Bower , so I've just separated into a repo, which works just the same as it used to do before, `bower install perfect-scrollbar`. I'll also update the repo with new versions for a while, in the way how React does, and users don't need to worry :)

DanielApt:
Happy to help with the Github Pages. What are the main things you want to achieve by renewing the Github Pages?

[...]

PR #293, opened 2015-03-02
#277 has been closed, remove reference to it in README.md

DanielApt:
No need to mention that perfect-scrollbar is quickly changing. v0.6.0 has been released, and I believe that this plugin won't change as quickly anymore.

Also issue no #277 has been closed, so no longer the need to mention it in README.md


hatashiro:
Fair enough. Actually I was about to do this. Thanks!

PR #300, opened 2015-03-08
Add wheelDisable option

ronihcohen:


hatashiro:
Instead of merging this, I may go with a more general disable option for several events, or custom callback in which the event can be prevented. I'll work on them some time.

Anyway, thank you very much for your interest!

PR #303, opened 2015-03-11
parseInt() returns NaN when the first argument is empty string.

damchoi:
ex) parseInt('', 10) -> NaN

### DIFF
--- a/src/js/lib/helper.js
+++ b/src/js/lib/helper.js
@@ -8,7 +8,7 @@ var cls = require('./class')
 
 exports.toInt = function (x) {
   if (typeof x === 'string') {
-    return parseInt(x, 10);
+    return parseInt(x, 10) || 0;
   } else {
     return ~~x;
   }


hatashiro:
Hi,

First of all, thank you for your interest on the plugin. However, I'm afraid I'm not sure why the result should be `0` when it gets an empty string. If it's an empty string which is surely not a proper integer string, the result should naturally be `NaN`.

Anyway, I'm sorry the PR can't be merged and thanks again.

Cheers,

damchoi:
Hi,

Actually, I've been encountering this issue in real works (I'm using the latest version of chrome). So this is not for "just in case" and my fix have solve the issue indeed.

Please reconsider the request.

hatashiro:
Hi,

I didn't mean that it's for "just in case". I meant if there is the issue you've been encountering, I'd fix the base reason that an empty string is indeed passed to the function, which shouldn't happen, rather than making it workaround.

Can you provide a stack trace of the problem for me to narrow down the reason why an empty string is passed? I'll personally look into the issue as well.

Cheers,

damchoi:
OK. I've attached some screenshots.

callstack_0
callstack_1

As you see, `window.getComputedStyle(element)[styleName]` (in `cssGet()` function) returns empty string when `styleName` is `marginRight`, `borderLeftWidth` or some.

hatashiro:
Oh, now I see. Cause `getComputedStyle` gets style values even though it's not been set, I wondered why it returns an empty string. But I found that if an element is detached from the DOM tree, it returns an empty string. I'll work on this and make a patch to cover the case.

Thanks for the report.

hatashiro:
4a7f27356afb7814698c3804615f8c8d6b224f8f has been uploaded and I hope this patch fixes the problem by ensuring there are proper scrollbars in the container. The patch will be included in the next release.

If there's any further problem, feel free to contact me :smile: 

Cheers,

damchoi:
I applied and tested the patch, unfortunately the issue still occurs. I will investigate more, and have plan for creating some example or test about this issue later.

Thank you for the support, though :)

PR #320, opened 2015-04-13
Support large and percentage margins on scrollbar rails

jaronkk:
We have a use case for a scrollbar that is smaller than the content area, but the drag and rail clicking worked incorrectly.

This pull request allows large rail margins to work as expected, and allows for percentage margins.

You can see my changes in examples/scrollbars-with-margins or at http://jaronkk.github.io/perfect-scrollbar/

Compare to the old functionality at http://jaronkk.github.io/perfect-scrollbar/original.html

This pull request does the following:
1. Calculates a ratio of rail size to container size. (railXRatio and railYRatio) and uses those values for drag and click detection.
2. Accounts for the window scroll position when clicking on a rail using window.scrollX and window.scrollY (to reproduce, go to http://noraesae.github.io/perfect-scrollbar/ , scroll down the page a bit, scroll the demo container to the bottom, then try clicking on the top part of the rail).
3. Fixed initial margin calculation to do allow for percentage margins.
4. Recalculates margin width and height on update, allowing for percentage margins.

### DIFF
--- a/examples/scrollbars-with-margin.html
+++ b/examples/scrollbars-with-margin.html
@@ -6,27 +6,39 @@
     <link href="../dist/css/perfect-scrollbar.css" rel="stylesheet">
     <script src="../dist/js/perfect-scrollbar.js"></script>
     <style>
-      .contentHolder { position:relative; margin:0px auto; padding:0px; width: 600px; height: 400px; overflow: auto; }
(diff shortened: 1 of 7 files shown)


hatashiro:
Looks quite handy. Now reviewing.
Thanks!

PR #321, opened 2015-04-15
changed toInt()'s function behavior

juxtapos:
When toInt() is called is with the result of getting a computed CSS style via the exported "css" function of the DOM module (e.g. as in the Instance prototype), that might return an empty string, it returned NaN. This caused some calculations to be broken, e.g. for the scrollbar's rail height. This simple fix will always return 0 instead of NaN and thus fixes scrollbar rails calculation.


hatashiro:
Actually, this is duplication of #303. For the PR, I rejected. However now I think it's better to move on.

Thanks!

PR #347, opened 2015-06-16
IE touch device fix

theiawolfe:
Add vendor specific css for IE to let javascript handle touch events instead of letting css attempt to handle them. Without this CSS property IE touch devices do not work. This does not affect IE mouse input behavior.


hatashiro:
Cool. Thanks!

Issue #357, opened 2015-07-06
TypeError: Cannot read property 'event' of undefined


adammiles10:
While using with angular 1.4.1 and ui-router.  Encountered the error when a view was getting replaced by children content.  My guess is that it was added to the dom, then removed before the proper destroy methods were called.  

The error is happening at the following line in the destroy.js file

i.event.unbindAll();

Possible Solution:
adding a null check before trying to unbind and remove.

I've created a fork and will make a pull request shortly.  Not sure if it is the most optimal solution but has worked for me.

Thanks,

Adam

noraesae:
Thank you for your report! I've just added a null check and error-throw like this: f9f20eeb6fb33bcab08dfe4868ebcf5668177973.

Your solution is good enough, but I thought it would be better to have an error message for debug later. But I really appreciate your report and solution upon the problem.

Thanks!

PR #358, opened 2015-07-06
Fix for undefined access in destory.js file

adammiles10:
Not sure if it is the best solution, but has worked for us.

in regarding https://github.com/noraesae/perfect-scrollbar/issues/357


hatashiro:
Fixed with f9f20eeb6fb33bcab08dfe4868ebcf5668177973. Thanks!

PR #359, opened 2015-07-07
Add null-check when updating or destroying an instance.

adammiles10:
Hey Jun,

(not sure if the best practice was to do another pull request) but here it is.  So I've found the issue is with the scroll bar being a direct child of an angular ui-router element.  I have added in the console.error() as you suggested.  But using console isn't supported by all browsers, (IE) especially, and not sure it's good to use here.  I'd suggest removing it, imo.

Thanks,

Adam

### DIFF
--- a/src/js/plugin/destroy.js
+++ b/src/js/plugin/destroy.js
@@ -10,11 +10,14 @@ var d = require('../lib/dom')
 module.exports = function (element) {
   var i = instances.get(element);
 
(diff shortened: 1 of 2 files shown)


hatashiro:
As you suggested, I just remove the line throwing an error and make it return early. Pls refer to 6c642d8a478ce231aad98aee7f6cd541cac816d2.

Thanks!

PR #361, opened 2015-07-08
add option to position the x rail to the bottom of the page

oste:
Hello @noraesae, I thought you might want to consider this pull request. It provides a new feature to fix the X rail to the bottom of the window as the user scrolls. Check out the new example file fixed-scroll-x.html. If it doesn't fit in your vision for the library I can understand but I think it is a nice feature that could be useful to others :)


brunowego:
+1

DanielApt:
Hi @oste,

First of all my apologies for getting back to you so late. Here's a well overdue response: I've reviewed your work, and although I think it's great I don't believe this ties in with the vision for perfect scrollbar.

Perfect scrollbar is trying to be as close to a native scrollbar as possible, and this behaviour is not common for native scrollbars.

Please don't let this dishearten you though, we have some great [issues](https://github.com/noraesae/perfect-scrollbar/issues) being reported which need addressing, and feel free to suggest any improvements you see fit yourself.

PR #366, opened 2015-07-23
Allow the clicking of a rail to propagate

dorilla:
There are times when I need the click event to propagate so that I can catch information about the event on my end. Default is true so that this change is optional.


hatashiro:
Great work! Thanks!

PR #373, opened 2015-08-05
0.6.4 btcc

WXLWEB:


hatashiro:
Hi,

I appreciate this PR! However, I'm afraid the patch for custom CSS may be used personally, and not merged into master.

Thanks!

PR #379, opened 2015-08-17
Add shift+space support for keyboard

srcn:
Hello @noraesae , native scrollbars have shift+space keymap for scrolling upwards so I thought it might be good to have that in perfect-scrollbar as well.


hatashiro:
Great work. Thanks!

Issue #94, opened 2014-01-24
Feature Request : Publish Scrolling Events on Container


WebStew:
Would it be possible to publish some custom events when the container is either being scrolled on the X or Y plane so that we are able to hook in.

DanielApt:
It'd be nice to not only trigger events for which axis is being scrolled, but in which direction. So we'd have events called `ps-scroll-up`, `ps-scroll-down`, etc.

The issue is that I can't bind to `scroll` event, because they don't bubble up, which is required for my scenario.

Also the gist doesn't work with PerfectScrollbar, as the library has changed a lot since then.

PR #386, opened 2015-09-23
Dispatch scroll events

DanielApt:
## Feature

Perfect scrollbar will dispatch custom events:
### ps-scroll-y

This event fires when the y-axis is scrolled in either direction.
### ps-scroll-x

This event fires when the x-axis is scrolled in either direction.
### ps-scroll-up

This event fires when scrolling upwards.
### ps-scroll-down

This event fires when scrolling downwards.
### ps-scroll-left

This event fires when scrolling to the left.
### ps-scroll-right

This event fires when scrolling to the right.
## Technical

All modifications of `scrollTop` and `scrollLeft` are done by the new `updateScroll()` function. This is also the function which detects the scroll direction, and dispatches the events. `updateScroll()` can be found in `/src/js/plugin/update-scroll.js`
## Improvements

This addresses issue no. #94

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,6 +312,41 @@ The number of pixels the content height can surpass the container height without
 When set to false, when clicking on a rail, the click event will be allowed to propagate.  
 **Default: true**
 
+## Events
+
+perfect-scrollbar dispatches custom events.
+
+### ps-scroll-y
+This event fires when the y-axis is scrolled in either direction.
+
+### ps-scroll-x
+This event fires when the x-axis is scrolled in either direction.
+
+### ps-scroll-up
+This event fires when scrolling upwards.
+
+### ps-scroll-down
+This event fires when scrolling downwards.
+
+### ps-scroll-left
+This event fires when scrolling to the left.
+
+### ps-scroll-right
+This event fires when scrolling to the right.
+
+You can listen to these events either with vanilla JavaScript
+```javascript
+document.addEventListener('ps-scroll-x', function () {
+  // ...
+}
+```
+or with jQuery
+```javascript
+$(document).on('ps-scroll-x', function () {
+  // ...
+}
+```
+
 ## Contribution
 
 #### Please read [Contributing](https://github.com/noraesae/perfect-scrollbar/wiki/Contributing) in the wiki before making any contribution.
--- a/README.md
+++ b/README.md
@@ -338,13 +338,13 @@ You can listen to these events either with vanilla JavaScript
 ```javascript
 document.addEventListener('ps-scroll-x', function () {
   // ...
-}
+})
 ```
 or with jQuery
 ```javascript
 $(document).on('ps-scroll-x', function () {
   // ...
-}
+})
 ```
 
 ## Contribution
--- a/examples/events.html
+++ b/examples/events.html
@@ -1,0 +1,70 @@
+<!DOCTYPE html>
+<html>
+<head>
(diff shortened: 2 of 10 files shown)


hatashiro:
Looks superb :thumbsup: Can you add an example too? If you're busy, I can do.

hatashiro:
And I'd like to add you as a contributor of this project, thanks to your wonderful helps. Please feel free to tell me if you like it.

README.md
```diff
@@ -312,6 +312,41 @@ The number of pixels the content height can surpass the container height without
...
+This event fires when scrolling to the left.
+
+### ps-scroll-right
+This event fires when scrolling to the right.
+
+You can listen to these events either with vanilla JavaScript
+```javascript
+document.addEventListener('ps-scroll-x', function () {
+  // ...
+}
```
hatashiro: Maybe

``` javascript
})
```

README.md
```diff
@@ -312,6 +312,41 @@ The number of pixels the content height can surpass the container height without
...
+```
+or with jQuery
+```javascript
+$(document).on('ps-scroll-x', function () {
+  // ...
+}
```
hatashiro:
```
})
```

DanielApt:
Sure thing, I'll make the changes to the documentation, and I'll also add an example.

I'm happy to be added as a contributor, thanks!

DanielApt:
@noraesae Fixed the documentation and added an example.

hatashiro:
Great :smile: I added you as a collaborator too :+1:

hatashiro:
From now on, please feel free to push any commit, resolve any issue and so on. I will also review the work and leave a comment.

Thanks!

DanielApt:
Thanks @noraesae, I actually realised there could also be the need for events when the user hits the top / bottom of the chart. I was initially thinking of calling these `ps-scroll-top` and `ps-scroll-bottom`, but this wouldn't be semantic for events on the x-axis.

Any suggestions for some better naming?

hatashiro:
Til now I've recommended user using the custom callback to handle the event, like [this](https://github.com/noraesae/perfect-scrollbar/blob/master/examples/reach-the-end-callback.html), but I think now we have a custom event interface and there's no reason not to implement that.

I'm also not sure about the event name, but may using `reach` in the names looks okay?

zomp:
You can call it `ps-scroll-start` and `ps-scroll-end` – it is used by CSS' [flexbox](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Using_CSS_flexible_boxes#Flexible_boxes_vocabulary).

DanielApt:
@zomp, good suggestion. I like calling these sides `start` and `end`. Now these events could occur both on y as well as the x axis.

Also `start` or `end` could imply something is starting/ending, but with the usage of the word `reach` this could be solved:

Bringing it all together I think we could use something like:
`ps-x-reach-start`, `ps-x-reach-end` (or `ps-x-reached-start`)?

zomp:
@DanielApt, I like the names. Infinitive `reach` is consistent with `click`, `load`, `focus` and other event names. The order (infinitive + specification) is consistent with `mouseover`, `keydown`...

DanielApt:
Alright, I'll implement the events `ps-x-reach-start`, `ps-x-reach-end`, `ps-y-reach-start` and `ps-y-reach-end`

DanielApt:
Just added the new events as well.

hatashiro:
Great work.

PR #414, opened 2015-11-04
Fix touch start not propagating issue

jinmingyougit:
Fix touch start not propagating issue, delegate this event to touchMove so it will still prevent parent container from scrolling.
Only need to trigger touch handler if the ps container is scrollable.
Need to get rid of global touch handler since it will set isGobalTouch = true and prevent touchmove handler from being executed.

### DIFF
--- a/src/js/plugin/handler/touch.js
+++ b/src/js/plugin/handler/touch.js
@@ -56,7 +56,7 @@ function bindTouchHandler(element, i, supportsTouch, supportsIePointer) {
   }
 
   function isScrollable() {
-    return (i.contentHeight !== i.containerHeight) && (i.contentWidth !== i.containerWidth)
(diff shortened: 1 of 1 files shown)


jmuerle:
I also found on my own that this change was necessary. Please merge this as soon as you can. Instead of adding the isScrollable function, though, I added a check into shouldHandle that returned false if both scrollbarXActive or scrollbarYActive were false.

DanielApt:
@jinmingyougit Yikes, sorry for the late response on our side in regards to reviewing your pull request.

I'll set time aside to review this issue this week.

DanielApt:
I've just looked a it, but I've not been able to understand what's currently not working correctly.

Could you please explain to me the situation (based off the current `master` branch) you are trying to fix, and how this pull request addresses it?

Once I understand I can implement your work. Thanks in advance.

hatashiro:
Closed for inactivity. Please feel free to reopen.

PR #422, opened 2015-11-22
Fix events not triggered when scrolling past boundaries

axelboc:
Instead of returning when scrolling past the container's boundaries,
override `value` to the max allowed and let the other custom events
fire if needed.

### DIFF
--- a/src/js/plugin/update-scroll.js
+++ b/src/js/plugin/update-scroll.js
@@ -43,29 +43,25 @@ module.exports = function (element, axis, value) {
   }
 
   if (axis === 'top' && value <= 0) {
-    element.scrollTop = 0;
+    element.scrollTop = value = 0; // don't allow negative scroll
     element.dispatchEvent(yStartEvent);
-    return; // don't allow negative scroll
   }
 
   if (axis === 'left' && value <= 0) {
-    element.scrollLeft = 0;
+    element.scrollLeft = value = 0; // don't allow negative scroll
     element.dispatchEvent(xStartEvent);
-    return; // don't allow negative scroll
   }
 
   var i = instances.get(element);
 
   if (axis === 'top' && value >= i.contentHeight - i.containerHeight) {
-    element.scrollTop = i.contentHeight - i.containerHeight;
+    element.scrollTop = value = i.contentHeight - i.containerHeight; // don't allow scroll past container
     element.dispatchEvent(yEndEvent);
-    return; // don't allow scroll past container
   }
 
   if (axis === 'left' && value >= i.contentWidth - i.containerWidth) {
-    element.scrollLeft = i.contentWidth - i.containerWidth;
+    element.scrollLeft = value = i.contentWidth - i.containerWidth; // don't allow scroll past container
     element.dispatchEvent(xEndEvent);
-    return; // don't allow scroll past container
   }
 
   if (!lastTop) {


DanielApt:
Hi there, I've just had the chance to review your code.

Just wanting to double check… you've named this PR _Fix events not triggered when scrolling past boundaries_.

When testing your code, events are actually _not_ triggered when scrolling past, instead of _being_ triggered, as your fix would imply.

In which scenario would your code cause events to be triggered, which in the past were not being triggered?

axelboc:
Hey, thanks for taking a look! :)

Sorry this wasn't clear in the commit message: I'm not referring to the `reach-(x|y)-(start|end)` events, but to the `scroll-(x|y)` and `scroll-(up|down|left|right)` events. Those `return` statements that I removed where preventing the other conditionals further down from being checked. Does that make sense?

DanielApt:
Oh, now I understand. So you're addressing a scenario where for example an upward scroll reaches the top.

In my previous code we'd only receive the event that it has reached the top, but thanks to your code we'll receive both that we reached the top, as well as that there was an upwards scroll.

Let me still double-check re. performance, as now we're going to go through the whole function, and previously we were only going through the function till we hit the first return.

axelboc:
Yeah that's it. No worries, please do!

DanielApt:
Thanks for the great work @axelboc. To my pleasant surprise I didn't notice any performance issues.

axelboc:
Happy new year @DanielApt! Thanks for merging this in. By any chance, would you have time to do another release some time soon?

DanielApt:
Apologies for the delay. I need @noraesae to do a release.

hatashiro:
Sorry for the late response. Will release a new version in a while.

axelboc:
Thanks! :)

zauni:
Just wanted to say, that this pull-request can change the order in which the events are firing. For example in this code:

``` javascript
$el = $('.foo');
$el.on('ps-x-reach-end', function() { $('.bar').addClass('reached-end'); });
$el.on('ps-scroll-x', function() { $('.bar').removeClass('reached-end'); });

scrollbar.initialize($el[0]);
```

When scrolling to the end, it was triggering the `ps-x-reach-end` event as the last event. But with this pull-request the `ps-scroll-x` event is triggerd as the last event.

In my case I want to add a class to another element if the scrollbar hits the end, and it should remove the class if the scrollbar is scrolling. But now the class is removed even if the scrollbar hits the end, because the scroll event is fired last.

Would it be possible to change the order of the events inside `update-scroll.js`, so that the `xEndEvent` and `yEndEvent` are fired last?

axelboc:
When `ps-x-reach-end` gets triggered, it means that the user has scrolled towards the right. So could you listen to `ps-scroll-left` instead of `ps-scroll-x`? By doing so, the class would be removed only when the user scrolls left. And you could do the same on the other side with `ps-x-reach-start` and `ps-scroll-right`.

zauni:
That's a good idea!

I solved it yesterday by saving the `scrollLeft` property on `reach-end` and checking in the `scroll-x` event handler if the `scrollLeft` property is different. But thanks for the suggestion :+1:

DanielApt:
I think @axelboc's suggestion is more elegant. Your logic should depend on the events that take place, and not per-se the order of the events.

PR #424, opened 2015-11-23
Fixes touch event handling in nested scrollbar containers

azala:
If you have one perfect-scrollbar container within another, on mobile devices, initiating a touch in the inner container, then scrolling beyond the bounds of the inner container, does nothing to the outer container even with swipePropagation enabled.

I'm not 100% sure why it works, but this hack fixes it.

### DIFF
--- a/src/js/plugin/handler/touch.js
+++ b/src/js/plugin/handler/touch.js
@@ -90,6 +90,9 @@ function bindTouchHandler(element, i, supportsTouch, supportsIePointer) {
     }
   }
   function touchMove(e) {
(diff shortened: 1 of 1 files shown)


DanielApt:
Just tested it, and can confirm the fix works.

DanielApt:
Thanks for the great work @azala!

azala:
@DanielApt No problem! Several mobile [Apples Against Humanity](https://azala.info) players had this problem after I switched to perfect-scrollbar, so I had to do something about it.

PR #434, opened 2015-12-10
replaced mixins with autoprefixer

wujekbogdan:
- added gulp-autoprefixer
- removed transition() and border-radius() mixins


DanielApt:
Thank you @wujekbogdan!

PR #441, opened 2016-01-14
IE 10 bug fix

alicjakujawa:
In IE 10 element.dataset return <div></div>, so it goes to else block and throw error - element.dataset.psId return undefined.

### DIFF
--- a/src/js/plugin/instances.js
+++ b/src/js/plugin/instances.js
@@ -79,7 +79,7 @@ function Instance(element) {
 }
 
 function getId(element) {
-  if (typeof element.dataset === 'undefined') {
+  if (Object.keys(element.dataset).length === 0) {
     return element.getAttribute('data-ps-id');
   } else {
     return element.dataset.psId;
@@ -87,7 +87,7 @@ function getId(element) {
 }
 
 function setId(element, id) {
-  if (typeof element.dataset === 'undefined') {
+  if (Object.keys(element.dataset).length === 0) {
     element.setAttribute('data-ps-id', id);
   } else {
     element.dataset.psId = id;
@@ -95,7 +95,7 @@ function setId(element, id) {
 }
 
 function removeId(element) {
-  if (typeof element.dataset === 'undefined') {
+  if (Object.keys(element.dataset).length === 0) {
     element.removeAttribute('data-ps-id');
   } else {
     delete element.dataset.psId;


DanielApt:
Could you create jsfiddle showing an error being thrown? I've not been able to reproduce the issue you're trying to fix.

Here's my jsfiddle https://jsfiddle.net/ggvu31vz/

I'd presume that the `destroy` method would hit the `getId` function, which would then throw an error. When testing in IE10 I however am not receiving this error.

DanielApt:
@alicjakujawa Could you please submit an [example with JSFiddle](https://github.com/noraesae/perfect-scrollbar#jsfiddle) demonstrating the issue?

tonypine:
I'm getting an error related to this to 

screen shot 2016-02-29 at 18 14 01

the setId and getId doesn't work as expected. This is an IE10.
I actually tested it on the JSFiddle that you @DanielApt referenced, and it worked, but 
there's a case that the getId n setId aren't working properly. I'm debugging here, but until now
i didn't find what's the cause of the bug. 

In my case, the solution of the @alicjakujawa worked.

DanielApt:
@tonypine could you try and create your own jsfiddle, or maybe upload a zip of your project?

Has this bug only happened when clicking the rail, or does it also happen with mousewheel, keyboard and other handlers?

tonypine:
@DanielApt yesterday with your JSFiddle and debugging I find out the reason that i'm getting that error. Maybe it isn't related to the one that @alicjakujawa are getting. I was using a [polyfill](https://gist.github.com/brettz9/4093766#file_html5_dataset.js) to improve the support to the dataset method, and that make the behavior different from that expected, that's the dataset to be `undefined`.

If you include this polyfill you should be able to re-create the bug. So don't know, if u want to extend the support to others using this polyfill you need to do a fix.

Look at this [JSFiddle](https://jsfiddle.net/tonypine/4o5un9ra/4/) simulating the bug. If you run it on IE10, u'll see a div with an `data-id`, with the polyfill you can access it with the `dataset.id`, so the dataset ins't `undefined`. But if u do, `dataset.id2 = 'test'`, it doesn't work, you can only set with the `setAttribute`, and after this you can get with the `dataset` method. So maybe you can change the setter to always use the `setAttribute` or improve the condition.

DanielApt:
I believe that @tonypine's PR fixes your issue @alicjakujawa. Feel free to re-open if you're still seeing issues in IE10.

PR #446, opened 2016-01-26
Update README.md

zedtux:
Add titles to the different way to install perfect-scrollbar and add the Rails gem link


DanielApt:
Absolutely stellar work @zedtux, I think this will greatly increase the usage of the library.

PR #449, opened 2016-01-29
Add parameter to give option of turning off scrolling with wheel. 

c4l3b:
Pamameter: useWheelScroll -- default: true

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,6 +304,10 @@ When set to true, and only one (vertical or horizontal) scrollbar is visible the
 When set to true, the scroll works with arrow keys on the keyboard. The element is scrolled only when the mouse cursor hovers the element.  
 **Default: true**
 
+### useWheelScroll
+When set to true, the scroll works with the wheel on the mouse or touch pad.
+**Default: true**
+
 ### suppressScrollX
 When set to true, the scroll bar in X axis will not be available, regardless of the content width.  
 **Default: false**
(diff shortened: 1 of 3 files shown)


DanielApt:
Thanks for the pull request @cmbertsch01 

Just so I understand, in what scenario would this behaviour be desirable?

c4l3b:
You're welcome and thanks for the consideration!
I love the perfect scrollbar project, and used it in a horizontal product
slider. It's very similar to the sliders on Amazon's homepage. There is an
issue, specifically when using a touch pad, when scrolling the whole page
and the cursor is over the perfect scrollbar container. It will catch any
horizontal scroll and stop scrolling the page. I wish I could give you an
example, but it would require you to be a registered user. Amazon's slider
is not bound to the wheel and I'm guessing it's for this reason.

Thanks!

Caleb

hatashiro:
I would like to implement this feature, but with a different form of option parameter. I'll do the work in short, but anyway will merge the patch for the time being.

Thanks!

hatashiro:
Oops, I found a little problem in this PR that it should be done in `initialize.js`, instead of the handler itself. I'll revert the patch. Sry, and thanks again!

hatashiro:
It's implemented with bd7134f33dbba3073d3f7f0d4204de8edf85ae72

PR #458, opened 2016-02-18
Fix stopPropagationOnClick handler

mingc00:
Adding native bind function as click event handler cannot stop event propagation.
Use e.stopPropagation()


hatashiro:
I reproduced the issue successfully with a fiddle: https://jsfiddle.net/yy9haq6h/2/

Thanks a lot!

Issue #455, opened 2016-02-12
Native scroll doesn't work inside perfectScrollbar


jormaechea:
Here's the fiddle: https://jsfiddle.net/9duy4tz8/

PR #459, opened 2016-02-18
Allow children native scroll

DEfusion:
If a child of an element with perfect scrollbar applied needed to scroll, e.g. by setting `overflow: scroll` then the mouse wheel event would be captured by the ancestor that has perfect scrollbar applied.

Detect these cases and don't capture nor handle the event on the perfect scrollbar ancestor.

Related to #455

### DIFF
--- a/examples/children-native-scroll.html
+++ b/examples/children-native-scroll.html
@@ -1,0 +1,94 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta http-equiv="content-type" content="text/html; charset=utf-8">
+    <title>perfect-scrollbar example</title>
+    <link href="../dist/css/perfect-scrollbar.css" rel="stylesheet">
+    <script src="../dist/js/perfect-scrollbar.js"></script>
+    <style>
+      .contentHolder { position:relative; margin:0px auto; padding:0px; width: 600px; height: 400px; overflow: auto; }
+      .contentHolder .content { background-image: url('./azusa.jpg'); width: 1280px; height: 720px; }
+      .my-list {
+        position: absolute;
+        top: 0;
+        left: 0;
+        width: 130px;
+        height: 200px;
+        overflow: scroll;
+        background-color: rgba(255,255,255,0.6);
+      }
+
+      .my-list-two {
+        left: 140px;
+        overflow: auto;
+      }
+
+      .my-list-three {
+        left: 290px;
+      }
+
+      ul {
+        margin: 0;
+        padding: 0;
+      }
+
+      li {
+        box-sizing: border-box;
+        padding: 10px 5px;
+        border-width: 0 0 1px 0;
+        list-style-type: none;
+        margin: 0;
+      }
+
+      li:nth-child(even) {
+        background-color: rgba(0,0,0,0.5);
+      }
+
+      .my-list-horizontal {
+        height: 75px;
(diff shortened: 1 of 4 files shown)


DanielApt:
Quickly looking at your work, this looks great. Will still briefly review on Sunday before merging it. Thanks in advance.

DanielApt:
Now I've had the chance to properly sit down and look at your pull request.

I sadly don't think I can accept this pull-request. I am happy to explain my—lengthy—reasoning, and _please_ correct me if my reasoning is wrong.

It seems that on every scroll [`shouldBeConsumedByChild()`](https://github.com/bandzoogle/perfect-scrollbar/commit/f376e0d0445a9d1e9b1e750eb38ef23e87c6096e#diff-e1daa77b9b9b3816651499d4e26a8cfeR95) will be called. This function then calls [`findScrollParent()`](https://github.com/bandzoogle/perfect-scrollbar/commit/f376e0d0445a9d1e9b1e750eb38ef23e87c6096e#diff-e1daa77b9b9b3816651499d4e26a8cfeR81). The issue lies with that second function.

If I understand correctly, it will check if the element is _either_ `event.currentTarget` _or_ if it has an `overflow` matching `scroll` or `auto`. If it matches neither of these conditions, it will repeat the same operation but for its parent element.

I think it is a rare scenario to have scrollable children within the `.ps-container`, which means that in most scenarios `findScrollParent()` will be called on _all_ elements between `event.target` and `event.currentTarget`. `findScrollParent()` will repeatedly call `window.getComputedStyle`, which can be a performance bottleneck ([as described here](https://gist.github.com/paulirish/5d52fb081b3570c81e3a#getcomputedstyle)). It is because of this performance bottleneck that I don't think I can accept this pull-request.

I do however agree issue #455 needs to be addressed, but here comes an alternative suggestion to address this issue:

A similar approach to [`shouldBeConsumedByTextarea()`](https://github.com/noraesae/perfect-scrollbar/blob/master/src/js/plugin/handler/mouse-wheel.js#L58) could be used. Instead of checking if `event.target` is a textarea, we could check if it has a classname (e.g. `.ps-child`), and is being hovered. 

To allow child elements to be scrolled, we could then simply check if the `.ps-container` has any children matching `.ps-child:hover`, if it does we allow `.ps-child` to be scrolled, otherwise we continue to scroll `.ps-container`

What do you think? Am I correct in my understanding that your pull-request would cause a performance issue, and would you agree that a better alternative would the `ps-child:hover` approach?

DEfusion:
Thanks for the update.

I also did consider using a classname for any children that will need to scroll but I thought not requiring that was easier for the end users and for it to just be handled out of the box. I did wonder about whether I should have added some caching so it only does the lookup on first scroll event, though I'm not sure how that would look.

However it's probably a bit of an edge case so maybe just specifying a class is just the best solution. I'm happy to re-work and re-submit a PR if you think that's the best way around.

DanielApt:
It would be absolutely brilliant if you resubmit a PR. Thank you in advance!

DanielApt:
Actually… I'll do it, as I actually want to combine `shouldBeConsumedByTextarea()` and standardise some logic.

DanielApt:
@DEfusion, finished the work on it yesterday. Thanks for your help, using your example helped me a lot!

PR #462, opened 2016-02-22
Fix for stopPropagation on IE

jonixj:
window.Event.prototype.stopPropagation does not exist in IE. 

cancelBubble can be used instead

### DIFF
--- a/src/js/plugin/handler/click-rail.js
+++ b/src/js/plugin/handler/click-rail.js
@@ -9,7 +9,14 @@ function bindClickRailHandler(element, i) {
   function pageOffset(el) {
     return el.getBoundingClientRect();
   }
-  var stopPropagation = window.Event.prototype.stopPropagation.bind;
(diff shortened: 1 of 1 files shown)


DanielApt:
Could you share an example demonstrating the current issue? This will allow us to review your work faster, and get your PR merged quicker.

hatashiro:
This may be an issue related with Event prototype in IE8. I'm afraid [we don't accept a patch for old IE browsers](https://github.com/noraesae/perfect-scrollbar#ie-support).

I really appreciate your contribution and sorry that we cannot merge this. If this issue's related even with modern browsers, please feel free to reopen.

hatashiro:
Or it may concern #458? The patch has been merged, so please test it out and tell me if there's any further problem.

jonixj:
Yes, sorry I did not see #458. That is a better solution for modern browsers.

PR #468, opened 2016-03-01
Use only getAttribute and setAttribute to maximize browser compatibility, due to partial support of the .dataset method

tonypine:
refer to #441 pull request
I think would be better to simply use getAttribute and setAttribute instead of dataset. It has a better performance (see this [jsperf](https://jsperf.com/dataset-vs-getattribute-and-setattribute/3)), and will keep support good for IE9 up and of course also to all others browsers.

### DIFF
--- a/src/js/plugin/instances.js
+++ b/src/js/plugin/instances.js
@@ -79,27 +79,15 @@ function Instance(element) {
 }
 
 function getId(element) {
-  if (typeof element.dataset === 'undefined') {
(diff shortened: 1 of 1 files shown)


hatashiro:
Looks like a proper point, as even using `dataset` seems slower. Any other opinion?

DanielApt:
Merge it

hatashiro:
Thanks!

PR #471, opened 2016-03-05
Use files for npm and move CONTRIBUTING.md

vsn4ik:


hatashiro:
Some changes look good. Here are the only things I want to keep:
- I think `CONTRIBUTING.md` is not specifically for github. It's definitely okay to keep it in the root. Other famous projects such as jQuery and React are doing the same.
- I think using `"` or `'` is completely up to users. W3C also supports using both. I personally prefer single quotation marks and there is no need to change it(I know there are examples written with `"` though). I may change them, but only when I really should. Sorry :smile: 

Other changes are perfect. Will checkout and modify a little bit.

hatashiro:
Merged manually: f20a8532ffe412d03068ca79c16f7cdfba19f206

Thanks!

vsn4ik:
Thanks for the answer!

PR #474, opened 2016-03-09
Always check parentNode before removing dom element

mingc00:
Prototype.js overrides HTMLElement.prototype.remove, but there is a defect in the implementation. It does not check parentNode before removing. Sometime it will cause js error when we try to destroy Ps.

Demo: https://jsfiddle.net/Lcm1kL17/

### DIFF
--- a/src/js/lib/dom.js
+++ b/src/js/lib/dom.js
@@ -66,12 +66,13 @@ DOM.matches = function (element, query) {
 };
 
 DOM.remove = function (element) {
+  if (!element.parentNode) {
+    return;
+  }
   if (typeof element.remove !== 'undefined') {
     element.remove();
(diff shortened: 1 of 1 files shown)


hatashiro:
First of all, thank you for the contribute.

I think this is an issue of Prototype.js, none of PS. A library shouldn't override `remove` of `Element.prototype`, as it is already implemented in Web APIs. Also, even though one is implemented as a polyfill, `parentNode` should be checked before removing. You can see the documentation [here](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove). I'm afraid that I think Prototype.js should fix the issue.

I'm sorry that I cannot merge this PR, and thanks again for your help.

Issue #13, opened 2013-03-03
Scrollbar scrolling too far/ percentage based heights/widths.


Myhka:
Hey first of all, thanks for the awesome plugin, it's perfect! It really makes scrollboxes look great. 
I'd like to use it in production, however I'm running into couple bugs, which I'll try to explain in detail here.

Firstly, perfectScrollbar doesn't seem to work when attempting to use it on a div with a percentage based height. The scrollbars just don't show up at all. I've also noticed that when using it on a div with a percentage based width, if the viewing screen changes widths, the x-scrollbar shows up even though there is nothing to scroll. Is there a way to only use perfectscrollbar on overflow-y?

Secondly, it's possible I might be writing something incorrectly, but in certain places when I use perfectscrollbar the y scrollbar keeps scrolling down with blank space in the div. However, when you grab the scrollbar with the mouse, it scrolls to where it should be (at the true bottom of the div).
Here's the code I'm using that produces this error:
```
#index.html:
   <div class="container">
   <div>
...content is rendered here...
    </div>
    </div>
#application.css:
 .container{
height: 400px;
width: 400px;
overflow: hidden;
position: absolute;
top: 0;
right: 0;
}
#application.js:
 $(document).on('click', '#render-button', function(e){
	$('.container').perfectScrollbar();
});

noraesae:
This may have been fixed with a recent scrolling logic change. Closed for inactivity.

PR #486, opened 2016-04-14
Rounding error fix when scrollbar is all the way down or right

tkhyn:
Hi there and thanks for perfect-scrollbar.

I noticed a bug when the scrollbar is in the far bottom / right position. Basically each time I scrolled further (with the bar or with the mouse wheel) it would move one more pixel. I found this problem was mentioned in issue #13, but it did not happen all the time, so I suspected rounding errors and found it was likely (see [here](http://stackoverflow.com/questions/11724556/how-to-get-the-maximum-scroll-position-of-a-page-precisely)).

A simple way to mitigate this (as most browsers don't support sub-pixel scroll values) is to test if the delta between the wanted scroll value and the current one is 1 or lower, and consider there should be no change when it's the case. Hence this PR.


DanielApt:
Looks great, thank you!

Issue #160, opened 2014-06-19
IE issue: Jumpy scrolling in IE


noraesae:
I confirmed this in the very latest IE too.
Sorry to IE users. This is a bug and there'll be some patches for this issue.
#56 and #118 are redirected to this issue.

Any contribution from anyone will be welcomed.

Thank you.

doronsever:
any news on this issue? this is an awesome plugin but i have to get it working on IE 11...

yoerivdm:
Hi, I stopped looking for a solution.

It is a known glitch in internet explorer ... the solution lies in
modifying the html structure which breaks with the philosophy of the
scrollbar. 

I switched to http://manos.malihu.gr/jquery-custom-content-scroller
since my current project is nearing completion and Internet Explorer is
not to be ignored for the target audience.

On 2015-01-13 15:55, doronsever wrote: 

> any news on this issue? this is an awesome plugin but i have to get it working on IE 11... 
> 
> --
> Reply to this email directly or view it on GitHub [1].

 

Links:
------
[1]
https://github.com/noraesae/perfect-scrollbar/issues/160#issuecomment-69755772

RamanVerma09:
Hi,

I have used perfect scrollbar in my kendo grid content on both axis. It is appearing fine but its flickering while scrolling. It seems like there is some delay between scrolling and updating the scrollbars positions, causing a jumpy effect.
Please reply.

Thanks

noraesae:
Hiya,

I'm really sorry for the inconvenience here. Because of my shortage of environment and time, the process hasn't be done quickly so far. I'll try to do it asap. Any help will be really appreciated.

Cheers,

larvanitis:
In case it is of any help https://www.modern.ie/en-us/virtualization-tools#downloads

noraesae:
Thanks for @larvanitis's suggestion, I installed IE VM. Now this issue is the highest priority, and I'm working on it. It's kinda tricky and I'm not sure if I can fix it very quickly, but I'll do my best.

Cheers guys!

noraesae:
Okay guys, I narrowed down the reason. Please refer to the following fiddle.

http://jsfiddle.net/js5hwjjt/

As you can see in the script, we change `scrollTop` and `top` properties at the same time. It means, the red square shouldn't blink, because both properties are updated in the same event cycle. As expected, in many major browsers such as Chrome, Safari or FF, it doesn't blink.

[...]

PR #500, opened 2016-05-18
Support position unset

### LINKED ISSUE #346: A workaround when position is unset.
I'm using this scrollbar for my nav-menu in wordpress. I have a list of menu items that expands beyond the nav container (also my ps-scrollbar container) and is displayed inline. I also have sub-menu and that submenu is located at the bottom of the my body container (outside of nav container). Now the only way to put it outside the nav container is to not put postion: relative on nav container since the sub menu is absolute and that will hide it if the nav container is relative.

Is there a workaround on this one? Here's a js fiddle on what I'm trying to achieve. Remove the position: relative on #nav-container and you'll see how I want it to look like.

http://jsfiddle.net/vanduzled/ndboyf6k/

yshterev:
Hello, 
This pull request resolves #346 as well as make the scroll work for the fixed tables by [salzerdesign.com](http://salzerdesign.com/test/fixedTable.html).

I think your current table demo does not work any IE but Edge.
ie11-win7

It works by setting X and Y scroll rails to `position:fixed` and positioning them over the container. It is intended as a fallback, and it will not work if you have 2 or more nested scrolling containers inside each other, because fixed of the child container cannot hide inside the parent container. I have tested most options on IE10+, Latest Chrome and FF and it seems to work.

The only changes that may affect the current state of the plugin are in `instances.js`. I have no idea how `i.isScrollbarXUsingBottom` and `i.isScrollbarYUsingRight` can be anything but `true`, and they are used in rail position calculation.


hatashiro:
Sorry for the late response. I'm not sure when this patch is needed. Can you explain a bit? Also, please don't remove trailing 2 spaces from `README.md`. They are for new lines.

yshterev:
Hi! There are two major points for supporting scroll rails with `position:fixed`, which allows the scroll to work without setting `position` on the container:
1. The case with the fixed header table. The scroll should work as a substitute for default browser scroll. Currently there is no way to use the scroll without a container with `position: relative`. The point here is that there are cases where default browser scroll works, and `position: relative` on the container breaks the layout, like the fixed header tables, so it would be good for the plugin to have some fallback.
2. I am not 100% sure as I have not spent enough time on this, but I think using `position:fixed` for scroll rails fixes IE issues described #160.

Unfortunately, I have experienced other issues with the scroll. I am using the scroll for a single page app. It has complex DOM and multiple instances of the scroll are used in web components. I had problems with:
1. This pull request needs more work. Sometimes rails positions are not calculated properly.
2. Nested scrolls do not work well. Could be because of the fixed rails positions.
3. The scroll is slowing performance on IE 10/11 + the jumping issue

I have decided to go with the default browser scroll and I do not plan to work on this one, unless you think this pull request has some potential.

`README.md` trailing spaces removed by mistake :(

hatashiro:
I had a look again.

1. Setting `position: fixed` doesn't simply fix IE problem, as PS' scrollbar logic itself depends on absolute positioning.
2. Hooking window's scroll event is much more performance-critical than local scroll hooking. I may not depend on it.
3. I may not add extra branch for no-position case.

I am really sorry but I may not accept this PR. I know that IE11 scrolling issue, but it should be solved differently.

Thanks again.

PR #525, opened 2016-06-23
On keydown, check for e.isDefaultPrevented() [jquery] and e.defaultPrevented [native]

howey-aus:
Event could be a native browser event, so we should also check for "defaultPrevented" if the jQuery props don't exist.


hatashiro:
Thanks!

Issue #457, opened 2016-02-17
Scrolling causes continuous browser repaints


Zenkit:
Hey,

first of all thank you very much for this plugin! I just noticed that if you enable "flash painting" in Chrome, you can see that scrolling inside a perfect scrollbar container causes a lot (!) of repaints. The normal scrollbar of the browser doesn't have that issue. 

Do you have an idea why this is happening? 

Cheers, 
Jesse

DanielApt:
There are plenty of reasons this could be happening, as described [here](https://gist.github.com/paulirish/5d52fb081b3570c81e3a#file-what-forces-layout-md).

Once I have the chance, this is something I'd like to look in to. I believe that there's a lot we can do to improve the rendering performance (and thus address the large amount of repaints). 

Things that I think will be especially useful to address this issue is to:

* use `transform` instead of `top` and `left` positioning
* wrapping the rendering logic inside a `requestAnimationFrame`

goldfire:
Any thoughts on what this would take? The performance issues are pretty much make this a non-starter.

noraesae:
PS has been using `top` and `left` for its simplicity and native scroll compatibility. It will be using them in the future too. Repaints are inevitable, so the best solution will be, as Daniel suggested, using `requestAnimationFrame` to reduce the count. The version 1.0 release will be using `requestAnimationFrame`, which is still in very early stage though.

If performance really suffers because of repaints, my suggestion is:

- Consider using native scrollbar. Please keep in mind that using custom scrollbar is such a bad practice for performance.
- Consider using other custom scrollbars using CSS transform to render the content. It may be difficult to use native JS scroll interface with them though. I cannot suggest any, as I never use, but I guess there are some quality libraries.

goldfire:
It isn't true to say that "repaints are inevitable." If things are done correctly there is no need for repaints at all in many cases. If you are re-writing anyway, I'm not sure why you wouldn't at least explore the possibility of something that could have a significant impact on performance (probably much more than just `requestAnimationFrame`).

noraesae:
Yes, I actually have tested it recently and found it was not true. As you said, repaint doesn't occur just by updating `top` or `left`(thank you for your report again!). I will definitely try avoiding the repaints when writing a new scroll logic.

rasmushaglund:
[...]

PR #532, opened 2016-07-15
Adding Ps.enable/disable to allow the Perfect-Scrollbar to be disabled temporarily and re-enabled later easily

lpetrov:
I needed a way to temporarily disable the scrolling on a PS container.
Calling .destroy and the re-initialising was causing the scroll area to go to the top (e.g. do a .scrollTop = 0), for some unknown reason, which ... even that was "workaroundable" I thought that it would be a lot more useful to just do a .preventDefault on the "wheel" event.

Then I'd found out that I can just add this into the Ps library, so that it would leave the scrollbars visible (can be hidden using the .ps-disabled class and some styles or styled as "disabled") and apparently this works best :)

So here is the MR, as following the contributing checklist:

Example of how this is supposed to work (the fiddle would complain for missing the Ps.enable/disable, which is added in this PR): https://jsfiddle.net/11sqnr1q/5/

A working example can be found in:
./examples/enable-disable.html

I've not tested the 'touch' handler and the 'jQuery' version of the plugin (despite that i'd exposed the enable/disable methods in the code).

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,17 @@ container.scrollTop = 0;
 Ps.update(container);
 ```
 
+If you want to temporarily disable (but leave visible) the scrollbar, use
+`disable` (and `enable` to re-enable it later).
+
+```javascript
+Ps.disable(container);
+```
+Enabling the scroller again can easily be done by calling:
+```javascript
+Ps.enable(container);
+```
+
 You can also get information about how to use the plugin
 from code in the `examples` directory of the source tree.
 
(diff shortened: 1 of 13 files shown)


lavrton:
Any chance to see the PR merged?
My use case is related to https://github.com/noraesae/perfect-scrollbar/issues/457. I am applying smooth animation scrolling to a container. Each `container.scrollLeft = value` produce browser repaints (because of PS style changes). For sure a better solution is to use `transform` instead of `top` and `left` properties.

hatashiro:
PS will have different API in v1. I really appreciate the work, but I am afraid I will adopt enable/disable after the new API is implemented.

I am really sorry that I do not merge this.

Issue #534, opened 2016-07-19
Calling .update from an ps-scroll-y triggers an InvalidStateError


lpetrov:
- [x] Check [FAQ](https://github.com/noraesae/perfect-scrollbar/wiki/FAQ)
- [x] Search if there's already one reported in Issues

From what I'd seen, this is most likely a duplicate of the:
https://github.com/noraesae/perfect-scrollbar/issues/505
(despite that ... this issue's comments/code were showing some extra/non-needed .update() calls in his example, so I'm reopening a new ticket, since his use case is just a bad/non-optimal usage of the library)
- [x] Prepare a JSFiddle reproducing the issue
- [x] Provide a page or source code where the issue can be checked

Demo: https://jsfiddle.net/bcyb7tmc/3/

<img width="1245" alt="screen shot 2016-07-19 at 3 31 19 pm" src="https://cloud.githubusercontent.com/assets/139877/16950032/c4d415e6-4dc6-11e6-91ca-28297b023c2c.png">

Basically, this is a simple "infinite scroll" use case.
The only difference (and what triggers the problem) is that the data is ready and available, so the .update is called in the same moment as the event handler is triggering the ps-scroll-y.
In a use case, where the data is loaded via ajax this would not happen...but for Single Page Apps (that most often have their data cached) this would be a major problem.

A potential solution is to re-create new instances of the events before every .dispatchEvent call.
I would try to prepare a PR with a proposed solution soon.

PR #535, opened 2016-07-19
Fixes https://github.com/noraesae/perfect-scrollbar/issues/534 (Calli…

lpetrov:
Fixes https://github.com/noraesae/perfect-scrollbar/issues/534.
Also, I'd added an example of an infinite scroll in examples/.


hatashiro:
Thanks!

PR #536, opened 2016-07-20
Defined dependencies for jspm

frg:
JSPM needs the dependencies defined in a particular format. If this is not done one would need to defined the dependencies manually. Also by not defining a jQuery version JSPM will use the version that is currently being used in the said project. JSPM will also handle the cases where jQuery is not being used just like any bundler.


hatashiro:
Looks great. Thanks!

PR #562, opened 2016-09-02
Added the "overflow: hidden" style in the container style example in …

aviklai:


DanielApt:
Thanks! Could you change it to `overflow: auto` instead?

aviklai:
Are you sure? it won't hide the native scroll in that case.

DanielApt:
`overflow: auto` is a graceful fallback. This makes sure that the element remains scrollable, even when there's no JS.

When JS is available, it applies the `ps-container` class, which forces the element to `overflow: hidden`.

Also perfect-scrollbar uses `auto`, for example [here](https://github.com/noraesae/perfect-scrollbar/blob/master/examples/events.html).

DanielApt:
If perfect-scrollbar is not hiding native scrolling when using `overflow: auto`, could you create an issue for this?

aviklai:
@DanielApt done. Issue number #566.

Anyway, it is not clear because in the Readme file, in the requirements section it is written that 
"the container must have an 'overflow:hidden' css style".

DanielApt:
Hi @aviklai,

Apologies to keep you waiting. I realised that adding `overflow: hidden` is not required. The `.ps-container` class already does this for you.

[Here's](https://github.com/aviklai/perfect-scrollbar/commit/73ff93984bc2aa14019b3998259b57b0c3ec9319) the commit which actually removed `overflow: hidden` requirement.

As a result. I will not be accepting your PR. Don't let that stop you. There are always [issues](https://github.com/noraesae/perfect-scrollbar/issues) to address or flag and we always need people like you to help! Once again, thank you.

PR #575, opened 2016-10-14
Mouse drag

jakkor:
Drag handler is for scrolling the container when we are dragging something in it.
For example when we drag something and are near the top border it should scroll up.
When we drag something near the bottom border it should scroll down.

This is really experimental for now, so not sure if you want a pull request. I had to write something like that for a project that I am working on. It seems to work, but needs a lot of improvements.

For Example:
- Better scroll speed. For now it's a fixed number (30pixels every 30 milliseconds). It should be something more fancy, like calculating from content size.
- only vertical scrolling is available for now, horizontal needs to be added
- when dragging something to different container scrolling does not work.
- This is more a prototype that production version. It needs much more testing and improvements.

I will probably work more on that feature but for now it may be useful for someone.

The handler name is 'mouse-drag' and it's disabled as default.


jakkor:
BTW I didn't add jsfiddle, but example is added to examples/options-handlers.html

hatashiro:
I appreciate your contribution. However, it's actually what `selection` scroll does. It is disabled by default because it's really buggy and creates more bugs. I'm sorry but I decided not to emulate this kind of scroll.

Thanks again.

jakkor:
I think that selection is a different thing as it for text selection ?
This is for reassigning elements in the element that is scrolled. For example when you have a list of divs and you want to reorder them by dragging. You start drag top element and you want to drag it to the bottom of the list. To do that you need to scroll the list. And you cannot scroll anything while you are dragging stuff.

This addition allows you to drag element to the bottom of the element and it will then start to scroll it.

Anyway I can use my forked version of this app if you don't want it here. Btw I already use this in production and it's working.

hatashiro:
Oh, you are right. Sorry for my misunderstanding.

I currently do not consider adding more scroll handler, as I think the current set is quite minimal and if another scroll is needed, the native scroll should be used. However, I will definitely take another look.

jakkor:
No worries, like I said I just use my own fork for that, so it's fine.

Kiruwagaka:
Any more info on this? This is exacly what I need.

Issue #578, opened 2016-10-22
missing "registry": "jspm" override in package.json's jspm section


tkhyn:
Hello,

When I run:

```
jspm install npm:perfect-scrollbar@0.6.13
```

this leads to:

```
err  Error on locate for github:jquery
GitHub packages must be of the form owner/repo.

err  Error locating github:jquery.

warn Installation changes not saved.
```

To fix it I have to use:

```
jspm install npm:perfect-scrollbar@0.6.13 -o "{registry: 'jspm'}"
```

So to make it work without having to provide an override, I'd suggest to add `"registry": "jspm"` in `package.json` as follows:

```
  "jspm": {
    "main": "./index.js",
    "registry": "jspm",
    "dependencies": {
      "jquery": "npm:jquery"
    }
  },
```

I can submit a PR if you don't see anything wrong with that change

noraesae:
Actually, I do not know much about jspm. Can anybody confirm this change?

tkhyn:
Maybe @frg, who authored 68dbc427ea8df2791472d8533a20d9ab01536650, can comment?

tkhyn:
As a side note, there were no jspm dependencies prior to 0.6.13 and that worked fine. @frg, what made you add the jspm dependencies in package.json?

Indeed, from what I know - or think I know (!) - jspm automatically parses the package code included in the package files (here `index.js` and `jquery.js`), even if there is a provided 'dependencies' override.

PR #591, opened 2016-12-03
Jquery is not a dependency

unsalted:
Jquery is not a dependency it is optional and should be treated as such.  Optionals will be part of 0.17 but that is still in beta.
Registry override will solve error #578


hatashiro:
Thanks. I have little knowledge of `jspm`. Could you update it when the optionals are released?

PR #595, opened 2016-12-19
Bind Touch or Pointer events exclusively, fixes #594

### LINKED ISSUE #594: Bug in Chrome 55: content jumps back a bit after scrolling with touch gesture
I have a Samsung Galaxy S7 Edge and recently noticed a bug there: when i scroll a menu down and after the scroll is done, content scrolls back a bit.
I've looked around the code, did some debugging and found that this is caused by easing functionality. Basically, it should continue scrolling, but in my case it did so in the opposite direction. After some further investigation it turned out to be caused by new Pointer Events addition to Chrome 55:
https://www.chromestatus.com/feature/4504699138998272
http://caniuse.com/#feat=pointer

Now touchMove event gets triggered twice and pageX, pageY values are different in each event case. I'm not sure why, but in the Pointer event they are rounded. Here is the result of scrolling straight down:
rsz_screenshot_from_2016-12-19_15-56-27
DifferenceY should always be negative, but because of rounding and other logic it becomes positive after Pointer event. And because it's triggered last, the easing function thinks we are scrolling up and eases up, when in fact it should ease further down.

RavWar:

### DIFF
--- a/src/js/plugin/handler/touch.js
+++ b/src/js/plugin/handler/touch.js
@@ -152,9 +152,7 @@ function bindTouchHandler(element, i, supportsTouch, supportsIePointer) {
     i.event.bind(element, 'touchstart', touchStart);
     i.event.bind(element, 'touchmove', touchMove);
     i.event.bind(element, 'touchend', touchEnd);
-  }
-
-  if (supportsIePointer) {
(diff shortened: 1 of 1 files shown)


hatashiro:
Looks great. Thanks!

Issue #600, opened 2017-01-09
adopting bem naming convention


loominade:
What do you think about adopting the [BEM naming convention](http://getbem.com/naming/)

It was already suggested in https://github.com/noraesae/perfect-scrollbar/issues/553#issuecomment-255194661.

The difference in the Dom should be a minor one:

```diff
- <div class="ps-container ps-theme-default ps-active-x ps-active-y">
+ <div class="ps ps--theme-default ps--active-x ps--active-y">
-   <div class="ps-scrollbar-x-rail">
+   <div class="ps__scrollbar-x-rail">
-     <div class="ps-scrollbar-x"></div>
+     <div class="ps__scrollbar-x"></div>
    </div>
-   <div class="ps-scrollbar-y-rail">
+   <div class="ps__scrollbar-y-rail">
-     <div class="ps-scrollbar-y">
+     <div class="ps__scrollbar-y">
    </div>
  </div>
```

## Block

In BEM methodology this would define the perfect scrollbar as a _block_ with the namespace `ps`.

## Elements

It has the following _elements_:

Element                | Description
------------------|-----------------------------------
`scrollbar-x-rail` | the rail on which X indicator is moving
`scrollbar-x`       | the indicator for the X coordinate of the scroll position
`scrollbar-y-rail` | the rail on which Y indicator is moving
`scrollbar-y`        | the indicator for the Y coordinate of the scroll position

All must be prefixed with the namespace `ps__`.

## Modifiers

The following _modifiers_ are possible.

Modifier                | Meaning
------------------|-----------------------------------
`theme-*`            | The theme of this particular instance of the the scrollbar. May be used for additional styling. The `*` may be replaced with any lowercase alphanumeric string that may also include single dashes (`-`)
`active-x`            | Indicates this scrollbar may be scrolled along the X axis
`active-y`            | Indicates this scrollbar may be scrolled along the Y axis
`x`                       | Indicates that `scrollbar-x` is currently dragged
`y`                       | Indicates that `scrollbar-y` is currently dragged

All must be prefixed with the namespace `ps--`.

noraesae:
Sounds good. I personally do not have any experience with BEM. Could you make a PR?

loominade:
> I personally do not have any experience with BEM.

I am myself not very experienced with it. It would be nice if someone from [github.com/bem](https://github.com/bem) could review it.

tadatuta:
[...]

PR #602, opened 2017-01-10
implemented BEM

loominade:
See https://github.com/noraesae/perfect-scrollbar/issues/600

### DIFF
--- a/src/js/plugin/instances.js
+++ b/src/js/plugin/instances.js
@@ -32,15 +32,15 @@ function Instance(element) {
   i.ownerDocument = element.ownerDocument || document;
 
   function focus() {
-    cls.add(element, 'ps-focus');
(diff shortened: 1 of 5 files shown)


src/js/plugin/instances.js
```diff
@@ -32,15 +32,15 @@ function Instance(element) {
   i.ownerDocument = element.ownerDocument || document;
 
   function focus() {
-    cls.add(element, 'ps-focus');
+    cls.add(element, 'ps--focus');
```
loominade, line 35: @noraesae does this class do something? I was not able to trigger it.

mhoyer:
Sidenote: I propose to def. mention the concrete changes of CSS class names with next release. In fact, isn't that change a breaking change?

We already saw broken custom styles in our application after updating to latest master for testing.

PR #603, opened 2017-01-10
Replaced obj.constructor with Array.isArray

### LINKED ISSUE #601: Helper.js - clone() isArray edge case
I'm using perfect-scrollbar in a project that works as an overlay on users' existing websites. Some of our users use Prototype JS which overrides the constructor for arrays. In `helper.js`, `clone()` does its array checking using `obj.constructor === Array` which in the case of Prototype will fail. However, using `Array.isArray(obj)` does work. `isArray` is supported back to IE9 according to MDN, but I'm not sure if there are other side effects.

This is a small edge case for a handful of our users, but it would save me a bit of trouble. I'd be happy to make a PR. Thanks!

hgascoigne:
Fixes #601 

This handles edge cases where a user is using something like PrototypeJS which overrides the default Array constructor which prevents perfect-scrollbar from working properly.

### DIFF
--- a/src/js/lib/helper.js
+++ b/src/js/lib/helper.js
@@ -10,7 +10,7 @@ var toInt = exports.toInt = function (x) {
 var clone = exports.clone = function (obj) {
   if (!obj) {
     return null;
-  } else if (obj.constructor === Array) {
+  } else if (Array.isArray(obj)) {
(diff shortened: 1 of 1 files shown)


hatashiro:
Looks nice. Thanks!

PR #609, opened 2017-01-27
Forcing propagation of wheel event to children

### LINKED ISSUE #577: Forcing propagation of wheel event to children
I ran into an issue with D3 and trying to capture a mouse wheel event to zoom on a chart where the event is not propagated properly. Using ps-child class is not enough, as the scrollTop logic returns false since the underlying child does not actually have any scrollable content.

My simple solution was to add an additional class that allows you to force consumption of the event by the child element:

`mouse-wheel.js -> shouldBeConsumedByChild`

```
if (child.className.match(/ps-must-propagate/)) {
    // force consumption by child
    return true;
}
```

Would you include this if I make a PR for it?

prog8:
Fixes #577

### DIFF
--- a/src/js/plugin/handler/mouse-wheel.js
+++ b/src/js/plugin/handler/mouse-wheel.js
@@ -62,6 +62,10 @@ function bindMouseWheelHandler(element, i) {
   function shouldBeConsumedByChild(deltaX, deltaY) {
     var child = element.querySelector('textarea:hover, select[multiple]:hover, .ps-child:hover');
     if (child) {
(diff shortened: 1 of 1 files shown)


hatashiro:
It should check `classList`, not `className`. However, it's fixed in the `next` branch.

- https://github.com/utatti/perfect-scrollbar/blob/next/src/handlers/mouse-wheel.js#L6
- https://github.com/utatti/perfect-scrollbar/wiki/FAQ#scrolling-children-inside-perfect-scrollbar

Thanks for your contribution!

Issue #607, opened 2017-01-18
Removing/Adding 'scrollbars-visible' class causes event 'ps-scroll-y' and 'ps-scroll-up' to trigger


Avuthless:
Hello,

This issue is really simple one. I add/remove 'scrollbars-visible' class from my components to hide scrollbars on demand. Unfortunately, when i enable/disable (add/remove) this class, events 'ps-scroll-up' and 'ps-scroll-y' are getting fired.
This is the code used to remove the scrollbars:
`this.element.nativeElement.classList.remove('scrollbars-visible');`

Unfortunately i cannot get any windows offset on events when using PS to check if it has really scrolled.

PR #618, opened 2017-02-13
Avoid unnecessary ps-scroll-x/y events

mhoyer:
I am a co-worker of @Avuthless and we fiddled with #607. 

## Our actual issue

We were listening to the `ps-scroll-y` event to hide a tooltip in case of scrolling. But, this approach collides with our call of `Ps.update()` which in turn raises `ps-scroll-y` (through `Ps.updateScoll()`) even if the use didn't really scroll.

The given fix is simply adding further conditions to only raise `ps-scroll-y` and `ps-scroll-x` if the scroll position really changed.

I am not sure if these missing conditions were on purpose. At least for us it solved the problems.

- [ ] Provide the scenario this PR will address(some JSFiddles will be perfect)
  - Perfect Scrollbar JSFiddle: https://jsfiddle.net/DanielApt/xv0rrxv3/
  - With jQuery: https://jsfiddle.net/DanielApt/gbfLazpx/

### DIFF
--- a/src/js/plugin/update-scroll.js
+++ b/src/js/plugin/update-scroll.js
@@ -60,11 +60,11 @@ module.exports = function (element, axis, value) {
     element.dispatchEvent(createDOMEvent('ps-x-reach-end'));
   }
 
-  if (!lastTop) {
+  if (lastTop === undefined) {
     lastTop = element.scrollTop;
   }
 
-  if (!lastLeft) {
+  if (lastLeft === undefined) {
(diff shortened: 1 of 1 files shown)


mhoyer:
I added further commits as we found another issue. There are `lastTop` and `lastLeft` defined in `update-scroll` to track last scroll positions. But it seems those variables are used globally which will conflict when having multiple instances of perfect scrollbar running.

PR #662, opened 2017-06-30
More logical event sending order

sconix:
When scrolling / swiping the events are sent in logical order (i.e. first comes the scroll events then the reach end events), but when setting the scroll position (through scrollTop for example) and calling update the event come in confusing order (first reach end event then scroll event).

This patch changes the order which the events are dispatched so that even in the latter case the events come in the order you would expect them to be.

If there was some logic for the events to be in the order they were then let me know, but at least this order makes more sense to me and removes extra checks for being at the end on scroll event from our application code. Since with current order when receiving scroll event we need to do some extra checks that are we still at end since we can't trust the reach end events.

### DIFF
--- a/src/js/plugin/update-scroll.js
+++ b/src/js/plugin/update-scroll.js
@@ -23,12 +23,10 @@ module.exports = function (element, axis, value) {
 
   if (axis === 'top' && value <= 0) {
     element.scrollTop = value = 0; // don't allow negative scroll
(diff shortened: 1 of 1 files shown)


hatashiro:
Thanks. I will copy diff into `next` manually, with some minor fixes.

sconix:
Thanks!

PR #679, opened 2017-08-31
Fixed tablets scrolling without touching/clicking

### LINKED ISSUE #619: Wacom tablet x Chrome - scrolls without clicking
This is a very specific scenario, but when using a Wacom tablet in Chrome, the scroll area moves as you hover over it, without clicking. Safari is fine, mouse is fine, trackpad is fine. It's just this Wacom x Chrome combination.

Any ideas? (Unfortunately I can't show a site to reproduce this, you need a wacom tablet and chrome to reproduce. But I'm seeing it on the main demo page here.)

TapGhoul:
Fixes #619

Pen pointers were previously allowing for scrolling even if no buttons were reported being pressed. Pens send a pointermove event every time they move, regardless of if any buttons are pressed.


hatashiro:
Thanks!

eelke:
@Silvea12 Thank you for this fix!
@noraesae The Github example needs an update to reflect this fix, right now the problem still occurs here. Can you update and bump the version?

hatashiro:
Sure. The patch has been released with v0.8.1.

eelke:
@noraesae Thank you, it works fine now!

TapGhoul:
We ran into this issue on Picarto.tv, we are using your scroll system. There is also one other weird bug but it keeps evading us whenever we test it. Will send a pull request when we find that.

Issue #695, opened 2017-10-16
Reach end/start events not send on update call with 1.x


sconix:
In 1.x it seems that the reach end/start events are not send on update call only when scrolling. By looking at the code the reach event sending is only send if the old position differs from the new position which is not the case when update is called in the init/manually. Is this a bug or intended behavior? 

I believe its a bug since the update code still has a comment '// Update top/left scroll to trigger events
' and as the updateScroll works now I don't see how it could send any events on manual update call.

If its intended behavior then is there any way to trigger those reach end/start events or check the status on init / after programmatic scroll without checking scroll position / geometry values manually.

sconix:
I do agree that this is better than how it was in the 0.8.x. I was kinda thinking more like a parameter for update / init so that the scroll state event could be send if wanted or providing a function to do the checking. I just hate "duplicating" a code so would be nice if the check done by the library would be available as helper function etc. Since its really common that you relay on the reach end event that you might want to know the status after init / programmatic scrolls as well.

utatti:
I agree that the new way can result in code duplication. I’ll see if I can add a helper.

utatti:
I've added a `reach` property, which always watching the state of scroll reach. With this you do not need to check scroll properties manually.

For its usage, please refer to `events.html`. I've updated README too.

- https://github.com/utatti/perfect-scrollbar/blob/master/examples/events.html#L72

It will be released in the next version. I'm not sure when it comes, maybe soon 😄

PR #701, opened 2017-10-20
Added updateReach and emitEvents parameters for the update method

sconix:
Regarding the issue #695, the added reach parameter helps a bit, but still forces having duplicating code for checking the reach state after programmatic scroll specially since the updateScroll call in update was removed.

My suggestion is in this PR, very simple changes to allow updating reach / scroll with update call and also to be able to send the events on if wanted. With these changes I would not need to replicate any of the perfect-scrollbar code in the ngx-perfect-scrollbar angular wrapper library.

I would really appreciate if you can consider merging this or discussing about better solution.


hatashiro:
I think we actually we do not need the parameters, as using them by default seems like making sense.

I think we can separate the task to make it default, so I'll merge this when it passes the CI.

sconix:
I can't say for sure which makes more sense having them as default or not, but imho having an update call with more finer control about what to update is not a bad thing. Anyway I am happy with either way.

hatashiro:
Please exclude `dist` from updates. They are updated only when a new version is released.

sconix:
Ah sorry, forgot that I run build so should not have used -a.

hatashiro:
Merged. Thanks! I'll write a follow-up patch in a while.

hatashiro:
Uploaded a follow up. Basically the patch makes it default. As you mentioned, there are many cases where contents are dynamically changed and reach states are updated according to the change. I guess it's completely natural to update `reach` and fire the reach events on `update()` call.

Thanks again! It will be released in the next version. I'm not sure when, but it'll be soon hopefully :)

sconix:
At least there hasn't been problems with 0.8.x which worked like that. If the need arises then its very easy to add the emitEvents parameter back, at least updating the reach values always should not be an issue to anyone. 

If you can release in few days it would be nice so I can get the next release out for the ngx-perfect-scrollbar which latest release I already had to pullback once since did not check carefully if the 0.8->1.x change broke anything :)

sconix:
Any change to ger new release out :) Angular 5 was just released so would like to release new major version of my wrapper and would love to have it use 1.x, but for that I need release with these event changes.

hatashiro:
Released into v1.2.0

PR #722, opened 2017-11-23
Fixed the issue with RTL where scrollbar continuously extends.

arl1nd:
I did run the "npm run format", hopefully this time is okay.


hatashiro:
Thanks!

hatashiro:
I actually reverted this, as it breaks usual cases.

arl1nd:
Sure no worries, I hope you can find a better alternative.

PR #723, opened 2017-11-25
Fix Scroll Propagation for Touchpads

konradjurk:
Original implementation caused choppy behavior when hitting container
bounds as it worked with delta===0 comparisons. This is perfectly fine
for wheels or browsers that do some sort of normalization (like Safari), but when `shouldPreventDefault` is feeded with raw touchpad data, moving the finger straight enough is unlikely, thus only a small percentage of scroll events were actually
propagated.

Improved implementation assumes the intended scroll axis by comparing
the absolute delta values of X and Y and propagates the events when
hitting the bounds on that axis.

Works after `npm format`, didn't know that 
  - Video footage https://youtu.be/cNneEmjcBzY
  - https://github.com/utatti/perfect-scrollbar/issues/639

### DIFF
--- a/src/handlers/mouse-wheel.js
+++ b/src/handlers/mouse-wheel.js
@@ -9,32 +9,21 @@ export default function(i) {
   let shouldPrevent = false;
 
   function shouldPreventDefault(deltaX, deltaY) {
-    const scrollTop = element.scrollTop;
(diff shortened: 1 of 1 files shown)


hatashiro:
Thanks!

PR #728, opened 2017-11-30
Add server-side rendering support

laurentgoudet:
This fixes https://github.com/utatti/perfect-scrollbar/issues/727.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ developers) scrollbar plugin.
 * Use plain `scrollTop` and `scrollLeft`
 * Scrollbar style is fully customizable
 * Efficient update on layout change
+* SSR friendly, e.g. through [ngx-perfect-scrollbar](https://github.com/zefoy/ngx-perfect-scrollbar)
 
 I hope you love it!
 
--- a/README.md
+++ b/README.md
@@ -17,7 +17,6 @@ developers) scrollbar plugin.
 * Use plain `scrollTop` and `scrollLeft`
 * Scrollbar style is fully customizable
 * Efficient update on layout change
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -17,6 +17,7 @@ developers) scrollbar plugin.
 * Use plain `scrollTop` and `scrollLeft`
 * Scrollbar style is fully customizable
 * Efficient update on layout change
+* SSR friendly, e.g. through [ngx-perfect-scrollbar](https://github.com/zefoy/ngx-perfect-scrollbar)
```
hatashiro: Supporting SSR is nice, but I'm afraid I personally do not think it's a selling point of ps. Can you please revert this?

src/lib/dom.js
```diff
@@ -4,12 +4,12 @@ export function div(className) {
   return div;
 }
 
-const elMatches =
-  Element.prototype.matches ||
-  Element.prototype.webkitMatchesSelector ||
-  Element.prototype.msMatchesSelector;
-
 export function matches(element, query) {
+  const elMatches =
```
hatashiro: It's better not to calculate `elMatches` every time `matches()` is called. Can you revert it to a root-level variable and just add a check for `Element`?

README.md
laurentgoudet: Fair enough, it's a selling point to me as since I've started to look into SSR I've realized it's really hard to find SSR friendly libraries :)

src/lib/dom.js
laurentgoudet: Good point, fixed it

hatashiro (approved):
Looks great. Thanks!

PR #730, opened 2017-12-04
Make `wheelPropagation` work on Right Container Bound

konradjurk:
When I enabled wheelPropagation on the demo page, I noticed that it works only on top, bottom and left container bounds. I confused `scrollWidth` with `offsetWidth` for `isRight` calculation. This PR fixes this.


hatashiro:
Thanks!

PR #731, opened 2017-12-04
Enable `wheelPropagation` on Demo Page

konradjurk:
As propagation is the default behavior in native scroll containers, I think
this will increase the perceived quality of the library when people try it out
on the Demo Page.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -113,7 +113,9 @@ button.addEventListener('click', () => {
     var heightInput = $('height');
     var button = $('button');
 
-    var ps = new PerfectScrollbar(demoDiv);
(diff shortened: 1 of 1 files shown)


hatashiro:
I think this is a valid point. Rather than setting it true only in demo page, it may be better to set it as default behaviour?

konradjurk:
@utatti I like that even better!

hatashiro:
I may still close this. Users may be confused when it works differently than how they see on demo page. Surely there is a source code, but I think demo better works with default behaviours. Sorry again, but please wait for #732. Thanks!

PR #732, opened 2017-12-05
Use `wheelPropagation` as default

konradjurk:


hatashiro:
This will break the default behaviour so I guess it should be released in the next major version. Thus I'm afraid it may take a while to merge this, but will definitely do. Thanks!

konradjurk:
Sure, but maybe we should add it to the demo page in the meantime so that the feature is demonstrated? Maybe I was too fast in closing that PR :D

hatashiro:
I will merge this now, and update README on the next release, which will be major.

PR #744, opened 2018-01-15
[Fix] fix rtl support

ahmad-sa3d:
Dear utatti,
Thank you for youe awesome library, i have encountered rtl issue and i have read issues #689 #713 #660
so i fixed this issue and now it's fully working as expected in both ltr, rtl
i followed your contribution guide and run tests

Thank Bro and i hope my contribution to this awesome library will be useful.

Issue #689, opened 2017-10-04
RTL support for Firefox and possibly MS browsers


utatti:
Currently, it seems RTL works only for Chrome. #484 

I'm not familiar with RTL and also do not own any Windows machine, so any contribution concerning this will be welcomed.

utatti:
I've confirmed that RTL support doesn't work in Chrome too, because of recent rail+thumb changes.

Exlord:
I am only using Y scroll so this changes worked for me , almost.

`.rtl .ps__rail-y {right : auto !important;left : 0;}`

```
let scrollbar   = new PerfectScrollbar(this.element, { suppressScrollX: false });
scrollbar.isRtl = false;
```

PR #750, opened 2018-01-30
Fix RTL Support (#689)

oladon:
Fixes rtl.html, and also adds a new example (examples/rtl-update.html) to show updating RTL after the initial creation of the scrollbar. Also tested in an actual use case which was previously infinite-scrolling to the right in Chrome.

Couple of items to call out:

- Inspired by #744 :)
- RTL felt more like a status than a sub-element for CSS purposes, but then I don't use BEM
- This does not include [line 84 of Ahmad's changes](https://github.com/utatti/perfect-scrollbar/pull/744/files#diff-9f8d3ecb2771fa202f724209c5d0fdb0R84), as I couldn't identify a need for that change (or in fact any behavioral difference with it)
- I'm not 100% comfortable with the magic number 9, but figured someone else more familiar with the project might be able to quickly identify it faster than me digging around :)

Please let me know if you have any questions.

- [✓] Read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [✓] Run `npm test` to make sure it formats and build successfully
- [✓] Provide the scenario this PR will address(some JSFiddles will be perfect)
  - [perfect-scrollbar JSFiddle](https://jsfiddle.net/utatti/dyvL31r6/)
- [✓] Refer to concerning issues if exist

### DIFF
--- a/css/perfect-scrollbar.css
+++ b/css/perfect-scrollbar.css
@@ -88,6 +88,11 @@
   position: absolute;
 }
 
+.ps--rtl > .ps__rail-y > .ps__thumb-y {
+  left: 2px;
+  right: unset;
+}
+
 .ps__rail-x:hover > .ps__thumb-x,
 .ps__rail-x:focus > .ps__thumb-x,
 .ps__rail-x.ps--clicking .ps__thumb-x {
--- a/examples/rtl-update.html
+++ b/examples/rtl-update.html
@@ -1,0 +1,58 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <link href="../css/perfect-scrollbar.css" rel="stylesheet">
+    <script src="../dist/perfect-scrollbar.js"></script>
+    <style>
+    #container {
+      position: relative;
+      margin: 0px auto;
+      padding: 0px;
+      width: 600px;
+      height: 400px;
+      overflow: auto;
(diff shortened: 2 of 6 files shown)


ahmad-sa3d:
thanks oladon for your updates, i think number 9 was fixing scrollbar width or position while i was fixing this issue on aproject i was working on, i donot know why but this was the fix, any way if it works will without it ok and it might be my project related, i will figure out with this new release and feedback

aghArdeshir:
Hey @oladon 
Thanks!
Everything is super-fine on RTL.
And in my non-multilingual rtl-only project I will use your version.

But the only thing that is annoying in your PR, is the example with dynamic `turn rtl on/off` which is great in chrome, but in firefox 66.0b12 (64-bit) it has buggy behavior which is depicted in gif below:
(thi gif is **5.9MB**, so it takes a little to load)

rtl-problem-on-firefox

oladon:
Hmm, that's odd behavior indeed. I normally use Firefox and haven't experienced it — do you have any browser extensions which might be interfering?

Unfortunately, not sure I'll have bandwidth to investigate in much detail any time soon.

Abdullah0991:
When this will be merged ?

mattonit:
This was already fixed in #744

PR #774, opened 2018-04-12
Mousewheel default deltamode values

stachoj:


stachoj:
Sorry, wrong repository. Will get back later, after proper testing.

PR #778, opened 2018-04-26
Fix demo page url

codemilli:
fix the demo page url from `github.com` to `github.io` which is a valid github page url.


hatashiro (approved):
Thanks!

PR #785, opened 2018-05-23
Corrected scrollLeft => scrollWidth

spoenemann:
While reading the code I found an asymmetry in mouse-wheel.js that looks like a bug. The maximal vertical scroll value is computed as  `cursor.scrollHeight - cursor.clientHeight`, but the maximal horizontal value is `cursor.scrollLeft - cursor.clientWidth`. I assume it should be `cursor.scrollWidth - cursor.clientWidth`.

### DIFF
--- a/src/handlers/mouse-wheel.js
+++ b/src/handlers/mouse-wheel.js
@@ -90,7 +90,7 @@ export default function(i) {
             return true;
           }
         }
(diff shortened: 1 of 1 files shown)


hatashiro (approved):
Yes, you're right. Thanks!

PR #788, opened 2018-05-29
Configurable swipe easing resistance

Schleuse:
I need the possibility to tune the resistance of the easing function, so I added an option to do that :smile:


Schleuse:
Hmm, I have no idea why the build is failing :disappointed:.
The tests run fine on my machine :man_shrugging:

PR #798, opened 2018-07-18
Make reversing axis with shift key to be executed only on windows (OS X supports natively)

darron1217:
  - [JSFiddle - change axis on shiftkey works well on both Mac & Windows](https://jsfiddle.net/dyvL31r6/3467/)
 - Issue mdbootstrap/perfect-scrollbar#104

Reverse axis on shift key is currently not working on OS X

OS X supports reversing natively, so they don't need the manual modification
So I changed the code to reverse delta only on Windows

### DIFF
--- a/src/handlers/mouse-wheel.js
+++ b/src/handlers/mouse-wheel.js
@@ -50,10 +50,14 @@ export default function(i) {
       deltaX = 0;
       deltaY = e.wheelDelta;
     }
-
-    if (e.shiftKey) {
-      // reverse axis with shift key
-      return [-deltaY, -deltaX];
+    
+    // Reverse axis with shift key should be executed only on windows
+    var ua = window.navigator.userAgent;
+    if (ua.indexOf('Windows') > 0) {
+      if (e.shiftKey) {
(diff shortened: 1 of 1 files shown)


sambaptista:
Using mac, I can confirm. On macOs, when holding shift key, e.wheelDeltaX and e.wheelDeltaY are reversed.

This lib reverses what is already reversed natively and cancels the horizontal scroll.

I've tested this fix and worked fine.

src/handlers/mouse-wheel.js
```diff
@@ -50,10 +50,14 @@ export default function(i) {
...
       deltaY = e.wheelDelta;
     }
-
-    if (e.shiftKey) {
-      // reverse axis with shift key
-      return [-deltaY, -deltaX];
+    
+    // Reverse axis with shift key should be executed only on windows
+    var ua = window.navigator.userAgent;
+    if (ua.indexOf('Windows') > 0) {
```
pierrebaronqumu, line 56: what about linux ? Shouldn't we check that it is not Mac instead ?

PR #837, opened 2018-12-05
Support fractional size of scrolled element

zimmicz:
Perfect scrollbar has issues working with fractional element witdth or height in IE and Edge as well. Try launching the [demo](https://utatti.github.io/perfect-scrollbar/) and setting height to 1000.5px and scroll the image. Even though it seems to fit the parent element, scrollbar becomes visible and can be scrolled all the way down.

PR should solve this issue by using `element.getBoundingClientRect()` for calculation.


pushpendersingh97:
When this is going to release?

vaibhavarora14:
I have the same question, 
@zimmicz , @utatti .
Do you have an idea when can this be included in official build?

Onyphlax:
Hi, this PR creates bug on Chrome / Firefox / Safari for the same issue, working with decimal value.
Easy to test : write 500.23px instead of 500px in examples/text-content.html line 32.

AmarMovva:
After upgrading to 1.5.0 in my application scrollbars appear where they should not suppose to be displayed since content perfectly fits and on start scrolling they disappear. this happens when running in Chrome and Firefox. I see large width difference from 1.4.0 to 1.5.0 which is causing scrollbars appear in places where they shoudn't. I thought `element.getBoundingClientRect().Width` returns `element.clientWidth` value in fractional but that is not the case always.

PR #943, opened 2020-04-24
Fix: unable zoom in mobile

zalesky:
If we have touches in different containers zoom isn't working.
Try to zoom when fingers at the different pictures here https://jsfiddle.net/od4j2tL3/

See also:
https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent/touches
https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent/targetTouches


zalesky:
Any comment?)

Issue #652, opened 2017-05-26
Passive event listeners


Hurtak:
Hi,

when I open chrome and have logging set to `Verbose` I am getting this warning:

```
[Violation] Added non-passive event listener to a scroll-blocking 'wheel' event. Consider marking event handler as 'passive' to make the page more responsive.
```

You can see it for example on this page http://noraesae.github.io/perfect-scrollbar/

I think it is worth the time to try if passive scroll listeners could work with this library.

Hurtak:
Oops, sorry duplicate of https://github.com/noraesae/perfect-scrollbar/issues/560

Issue #560, opened 2016-08-30
"passive" event support


mankins:
I've been noticing warnings in Chrome:

> "Handling of 'wheel' input event was delayed for 3198 ms due to main thread being busy. Consider marking event handler as 'passive' to make the page more responive."

A little more googling presented the "passive" event handler, recently added by browsers:

http://stackoverflow.com/questions/39152877/consider-marking-event-handler-as-passive-to-make-the-page-more-responive/39187679#39187679

And so I wonder if there's any advice on how to setup perfect-scrollbar to take advantage of this feature?

noraesae:
Interesting. For the time being, I have no idea about the passive event handler, but I will absolutely look into this, at least before working on v1.0.

mankins:
Thanks!

xDeepx:
I have same problem in Fuse http://fuse-angular-material.withinpixels.com/ui/forms

![image](https://cloud.githubusercontent.com/assets/16268668/19599639/d37364f4-97b2-11e6-8c2d-1f33eeec0b17.png)

![image](https://cloud.githubusercontent.com/assets/16268668/19599702/121ae0b0-97b3-11e6-8b3a-fcdb4e4d52ca.png)

lekhnath:
Any progress on it??

noraesae:
I've looked into the feature. Shortly, it is difficult to use it for the time being. Using passive event prevent scrolling from being blocked by main thread, by ensuring `preventDefault()` is not called. However, PS uses `preventDefault()` in the events for checking whether and how it should scroll. It may be possible in the future by modifying the logic and removing the predicates from PS, but at least not now.

Besides the passive event support, you can easily remove the warnings. Using passive events isn't the only solution for the problem. The warning means that scroll events are delayed because of heavy job in main threads. Thus, the problem is actually the heavy job in main threads, not the scroll events themselves. To fix it, just remove them and make the page light. If it is difficult to make them lighter but the problem still exist, please consider using native scrollbars. Keep in mind that using custom scrollbar is bad practice for performance.

I am afraid I close this issue here. If there is any good suggestion or implementation about this issue, please reopen it or upload a PR.

kinkoazc:
[...]

PR #945, opened 2020-05-01
passive events syntax update with feature detection

andrei-gheorghiu:
I read contributing. 
For some reason, I can't get either `format` or `format:lint` scripts to run on my machine. Complains about not matching any files.

If there are any linting errors, feel free to update it with linting according to your standards.
What I've done is quite simple: I've taken the feature detection from [here](https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md#feature-detection) and applied it to `EventElement.bind`.

If the browser supports passive, instead of `false` I'm passing `{ passive: false }`. 

Note the warning doesn't require not to call `.preventDefault()`. It only requires you specify passive one way or another. Since you have cases where you prevent default, passive should be false.

#560 
#652


andrei-gheorghiu:
I have no idea how that `Math.ceil` changed to `Math.round` I can only guess it's a change between your `1.4.*` and `1.5.*`. 

Not very experienced with contributions. Just let me know if you need me to do anything else.

Ivan-Parushev:
Can you check why pipeline is failing? It seems to be connected to the linting step

wilkie:
Hello! This is because the "single quote" option is being used and it also does not like the length of the lines. Picky little bugger.

Here is the commit to pull that does not irritate the linter: https://github.com/wilkie/perfect-scrollbar/commit/cc07f078c5b1a029e8b9bbb950b6731c44b39c7c

andrei-gheorghiu:
> Hello! This is because the "single quote" option is being used and it also does not like the length of the lines. Picky little bugger.
> 
> Here is the commit to pull that does not irritate the linter: [wilkie@cc07f07](https://github.com/wilkie/perfect-scrollbar/commit/cc07f078c5b1a029e8b9bbb950b6731c44b39c7c)

I pushed the suggested linting but apparently there's more to it. I also tried to downgrade my npm and node versions but it still fails at `npm run format:lint`, with: 

`[error] No matching files. Patterns tried: 'src/**/*.js' !**/node_modules/** !./node_modules/** !**/.{git,svn,hg}/** !./.{git,s
vn,hg}/**`

andrei-gheorghiu:
Have I merged into the wrong branch?

src/lib/event-manager.js
```diff
@@ -20,15 +20,23 @@ class EventElement {
       this.handlers[eventName] = [];
     }
     this.handlers[eventName].push(handler);
-    this.element.addEventListener(eventName, handler, supportsPassive ? { passive: false } : false);
+    this.element.addEventListener(
+      eventName,
+      handler,
+      supportsPassive ? { passive: false } : false
```
iomariani, line 26: Add a comma to the end to avoid linting error for [trailing comma](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas).

src/lib/event-manager.js
```diff
@@ -20,15 +20,23 @@ class EventElement {
...
   unbind(eventName, target) {
     this.handlers[eventName] = this.handlers[eventName].filter(handler => {
       if (target && handler !== target) {
         return true;
       }
-      this.element.removeEventListener(eventName, handler, supportsPassive ? { passive: false } : false);
+      this.element.removeEventListener(
+        eventName,
+        handler,
+        supportsPassive ? { passive: false } : false
```
iomariani, line 38: This line as well

iomariani (changes requested):

src/lib/event-manager.js
iomariani:
```js
supportsPassive ? { passive: false } : false,
```

iomariani:
I also came across this warning on my app...
Your fix seems to solve it.

For the linting error on the check, I've made notes on the commit.

andrei-gheorghiu:
> I also came across this warning on my app...
> Your fix seems to solve it.
> 
> For the linting error on the check, I've made notes on the commit.

Added, same result.

As a sidenote: perhaps we could improve the definition around steps to install everything required to run the project and the tests. 
I've installed node 5 (via nvm), removed all packages and ran `npm i`. 
I'm still getting `[error] No matching files. Patterns tried: ... a bunch of patterns`. Relevant error line:
`Failed at the perfect-scrollbar@1.5.0 format:lint script 'prettier -l --single-quote --trailing-comma es5 'src/**/*.js''`

Another side note, when installing all packages from scratch, I got an unmet peer dependency, at:

    -- rollup-plugin-buble@0.19.8
      -- buble@0.19.8
       +-- UNMET PEER DEPENDENCY acorn@6.4.1

Could this be the culprit?
____

Aside from all that: I'd love if anyone could just commit or PR the right linting for my changes, on my branch, so that we could get this over with. I've been using my branch successfully for a couple of months now and I'm not seeing any nasty warnings from Chrome while everything works as expected. All I wanted was to share the fix.

Cheers!

src/lib/event-manager.js
andrei-gheorghiu: Added, same result :(

andrei-gheorghiu:
I was able to figure out the error. Basically, prettier [doesn't recognize globs on windows machines](https://github.com/prettier/prettier/issues/4804), if using single quotes.

So I had to replace  `"... 'src/**/*.js'"` with `"... \"src/**/*.js\""` in both `format` scripts (now they work on windows machines as well).
For obvious reasons, I also pushed the changes to the script syntax (I'm guessing you want your `format` scripts to run on windows machines, right)? 
___
@iomariani looks like the missing comma was after the getter in `opts`, not in the function call.

jeffshaver:
I don't think this change works as expected. When using passive event listeners, I don't think you are allowed to call event.preventDefault. But this library does in the mouseWheel handler.

https://github.com/mdbootstrap/perfect-scrollbar/blob/ea2f8e625aa49f5af58e3a964953c1cb16853782/dist/perfect-scrollbar.esm.js#L900

andrei-gheorghiu:
> I don't think this change works as expected. When using passive event listeners, I don't think you are allowed to call event.preventDefault. But this library does in the mouseWheel handler.
> 
> https://github.com/mdbootstrap/perfect-scrollbar/blob/ea2f8e625aa49f5af58e3a964953c1cb16853782/dist/perfect-scrollbar.esm.js#L900

This tells me two things: 

 - you haven't actually looked at the code
 - you haven't actually tested it
______

The proposed pull request does not change the behavior of the library. It only changes the syntax from 

    this.element.addEventListener(eventName, handler, false);

to: 

    this.element.addEventListener(eventName, handler, { passive: false });

for browsers which support the new syntax. For legacy browsers, the syntax remains the same. 

The warning in modern browsers (i.e: latest webkit based)  is gone and the functionality is unchanged. 

A common misconception about the Chrome warning is that you have to pass in `{ passive: true }`. You don't. You have to pass any boolean value to `passive`. Passing `true` does disable the ability to `preventDefault()` on the target event, but that's not the case here, since `passive` is set to `false`, since we need to trigger `preventDefault()` in some cases.

jeffshaver:
@andrei-gheorghiu sorry. Read the code. But because of my preconceived idea of what this was (found this because I wanted to find issues/prs on the repo to enable passive event listeners instead of making a new one), I misread the { passive: false } for true.

I wasn't trying to come off in a bad way with my comment. I tried to make the changes in this or in my own copy of it but I put true instead of false and got errors in chrome. Was just trying to point that out. Apologies.

Splaktar:
Thank you for this PR!

I understand that this solves the warnings in Chrome and Lighthouse. But what is the performance impact of using this library if it can't use passive event listeners?

andrei-gheorghiu:
> Thank you for this PR!
> 
> I understand that this solves the warnings in Chrome and Lighthouse. But what is the performance impact of using this library if it can't use passive event listeners?

@Splaktar. From my POV, your question is off-topic here. But frankly, I don't really know where it _would_ be on-topic. And I think it's a fairly good question/concern. Since you asked it here, I'll answer it here. To everyone else, apologies for the _off-topic_. 

I'll start by saying I use perfect scrollbar a lot. I've used it in Angular and Vue projects. I also know its limitations as I've tested it extensively on a wide range of device/browser combos, while trying to figure out which would be the best solution for the problem this package is trying to solve.

The goal of the package is to replace native scroll mechanism with a fancier looking one, as the native desktop scrollbars in webkit browsers (which are by far the most popular) have their fair share of design shortcomings. Technically, the only way to achieve it is by calling `preventDefault()` on the native `scroll` event, changing the element's scroll-based props directly while also painting and animating a fancy scrollbar replacement. 
In short, there is no technical way of achieving the library's goal without canceling the native `scroll` event, which is to call `.preventDefault()` on it, which can only be performed by setting `passive: false` on the listener. Without `passive:false` on the listener, you can't replace the scrollbar. It's that simple.

In terms of performance, I think the code is quite performant. To be more exact, it does its job properly with the least amount of computation, and it's also using DOM techniques which do not trigger unnecessary repaints.

However, I believe it's a bit too ambitious. In my personal opinion, it should have a built-in option to only replace the scroll mechanism on desktop devices, not on every device. And it doesn't. 
Unfortunately, replacing the scroll mechanism on any touch based device is not a good idea, as touch devices are being unfair towards page scripts. In short, a significant range (all iStuff devices and some Android ones) are prioritizing scroll smoothness and, basically, they don't always declare the scroll event as it happens. Which means no matter how performant the library's code is, it will still be jumpy at times for the sole reason it doesn't get consistent updates from the device. Whenever the event is not passed, scrolling won't be perfectly smooth. While totally unfair, I can't really blame the device manufacturers, either (actually, I can - I blame mostly Apple, as they were the first to cheat like this - then, understandably, others followed). Anyway, blaming manufacturers is completely futile. I do understand their reasoning: they are sheltering native scroll from being slowed down by page scripts so their device is perceived as smooth and performant. Totally understandable in a competitive market, especially since the vast majority of users are not tech savvy and judge a device by how smooth it "feels" and "animates". Most times, that means "how smoothly it scrolls".

Based on all of the above, I have concluded the best possible experience is to only use perfect-scrollbar on desktop (pointer-based) devices and completely disable it on touch-based devices. Since the library doesn't provide this option out of the box, I have to use a dynamic template for the actual content in conjunction with a very tiny script detecting whether I'm on a touch device or not (`ismobile.js`). I then wrap the template in a `perfect-scrollbar` container (or not) depending on current device. Works really well, as the desktop browser scroll API is reliable and touch devices get their smooth native scrolling. The technique is a breeze in Vue, thanks to `<slot />`s. In Angular one has to get a bit dirtier and a bit more verbose. But hey, if you're an Angular consumer you probably like it a bit dirtier and a lil' (more) verbose. Kidding, of course. Or am I? :)

To conclude: I can't really fault the performance of the package, but it is unfairly faulted on touch devices by the devices themselves, as any other library replacing native scroll would be. The better way to counter it, IMHO, and to provide the best possible experience for your users is to enable perfect-scrollbar only on pointer based devices. 
But that has nothing to do with this PR.

jtraulle:
Hello :wave: 

Any status about merging this PR ? It has been a long time.

ridvanonal:
Any news on this issue? We are still waiting for this issue to be resolved. Thank you.

PR #948, opened 2020-05-22
Fix scrolling with useBothWheelAxes when both scrollbars are active

mciesin:
Scrolling with mousewheel doesn't work when using `useBothWheelAxes` with two active scrollbars.

  - [perfect-scrollbar JSFiddle](https://jsfiddle.net/spaknjgq/)


mciesin:
@filipkappa Why this is closed? The issue still exists, and the case when `useBothWheelAxes` is set to `true` and both vertical and horizontal scrollbars are active is not handled.

PR #969, opened 2020-08-28
Fix for TypeError: Attempted to assign to readonly property (fixes #968)

### LINKED ISSUE #968: TypeError: Attempted to assign to readonly property.
We are seeing `TypeError: Attempted to assign to readonly property.` in our error reporting at the location `perfect-scrollbar.js:6:3736` on some Safari browsers running on MAC OS. (We are using perfect-scrollbar 1.5.0)

I think this is, because in `drag_thumb.js` the readonly property `pageX` (or `pageY`) of the mouse event is tried to be set when the condition `touchMode && e.touches` is truthy.

flundermicha:


flundermicha:
ping

PR #1074, opened 2022-09-27
Added instructions for setting up component & site rendering using Node & EJS

artslimedev:
This is a pull request to add instructions to the README.md on how to get Node to be able to render EJS pages that include the Javascript some of the components require. The hope is that if this gets merged, these instructions get added to the website as well. While the fix is very simple, it may not come across as obvious to beginners who are learning to code and may be having their first framework experience with Tailwind Libraries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,6 +312,45 @@ Alternatively, you can import it in the following way (bundler version):
 
 ```
 import 'tw-elements';
+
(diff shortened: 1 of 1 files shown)


Trochonovitz:
Hi, we are going to prepare separately integration tutorial for Express.js (with EJS) so these instructions will be at the different place. However thank you for contribution!

PR #1118, opened 2022-11-25
fix dropdown animation

juujisai:


src/js/bs/src/dropdown.js
```diff
@@ -114,6 +115,7 @@ class Dropdown extends BaseComponent {
     this._config = this._getConfig(config);
     this._menu = this._getMenuElement();
     this._inNavbar = this._detectNavbar();
+    this._FADE_OUT_ANIMATE = null;
```
smolenski-mikolaj: Please use `_camelCase`

PR #1159, opened 2022-12-15
Add ripple effect to docs

### LINKED ISSUE #1127: Add default ripple effect with data-attributes

iprzybysz:
Added ripple effect to docs


site/content/docs/standard/designblocks/admin/index.html
```diff
@@ -21,7 +21,7 @@
     <div class="mb-12">
       <div
         class="relative overflow-hidden bg-no-repeat bg-cover ripple shadow-lg rounded-lg mb-6 flex items-center"
-        data-mdb-ripple-color="true" data-mdb-ripple-color="light"
+        data-te-ripple-init
```
smolenski-mikolaj, line 24: Is it an intentional omission of the `data-te-ripple-color="light"` attribute in design blocks?

iprzybysz: Yes, it is. I decided to add only `data-te-ripple-init` because images are bright and the ripple effect in light colour was invisible.

PR #1265, opened 2023-01-20
Timepicker - class customization

### LINKED ISSUE #1232: Timepicker - add class customization

iprzybysz:


src/js/mdb/timepicker/index.js
```diff
@@ -973,7 +963,7 @@ class Timepicker {
           e.target.blur();
 
           div.innerHTML = getTimepickerTemplate(this._options);
-          div.classList.add(...CLASS_MODAL);
+          div.classList.add(this._classes.modal);
```
smolenski-mikolaj: It has to be with with spread operator because there is an option that user will want to inject multiple classes.

src/js/mdb/timepicker/index.js
```diff
@@ -995,7 +985,7 @@ class Timepicker {
           if (this._animations) {
             this._toggleBackdropAnimation();
           } else {
-            this._wrapper.classList.add(...CLASS_OPACITY);
+            this._wrapper.classList.add(this._classes.opacity);
```
smolenski-mikolaj: same here

src/js/mdb/timepicker/index.js
```diff
@@ -1009,7 +999,7 @@ class Timepicker {
             );
 
             if (format12) {
-              this._PM.classList.add(...CLASS_OPACITY);
+              this._PM.classList.add(this._classes.opacity);
```
smolenski-mikolaj: and here

PR #1267, opened 2023-01-21
timepicker - Fix formatToAmPm bug

vincentinttsh:
bug => time will always be AM when hours is 12
EX: input `12:34 PM`


smolenski-mikolaj:
Thanks for contribution!

Unfortunately, the completely rebuilt Timepicker is already waiting for beta-1.0.0 to be deployed. If the problem persists in this version, please make a new pull request or issue.

Issue #1270, opened 2023-01-24
Default checked radio - not checked


flowoow:
I'm trying to use the Radio input and have a default selection to one of the radio. However, the radio input is not checked.

From the documentation: https://tailwind-elements.com/docs/standard/forms/radio/
The Basic example has the same issue.

Default checked radio is not checked

vinaysanga:
Fixed this issue in my pull request #1276

smolenski-mikolaj:
We've just release the new v1.0.0-beta1 version so after update the issue should be fixed.

PR #1276, opened 2023-01-26
Fix the grammar in contribute page, fix issue #1270

vinaysanga:
Fix the grammar in contribute page
- Modify the description so that the verb agrees with the subject and the meaning is well-established. 
- Modify the `<p>` tag so that the pronoun is correct.

Fix issue #1270
- Close the `<input>` tag properly.


smolenski-mikolaj:
Thanks for contribution, but unfortunately the fix for forms is already waiting for a release => https://github.com/mdbootstrap/Tailwind-Elements/blob/dev/site/content/docs/standard/forms/radio/index.html

Also, we will be updating the contribution page soon, so the content will probably change.

PR #1503, opened 2023-03-21
Change the way the theme switches

kpietron:


site/config/dark/config.toml
catchmareck, line 1: Since this is a copy of the `production` config but enables the `dark` theme, I would rename the file to `production-dark` in order to better express what it is and to avoid potential conflicts with test environments in the future.

catchmareck (changes requested):
Please see the comment about the config directory name

catchmareck (approved):
OK, now it looks good to me ;)

Issue #1641, opened 2023-05-08
Docs - section links in API tab won't open in a new tab, because they do not exist in Overview tab


juujisai:

juujisai:
https://github.com/mdbootstrap/TW-Elements/pull/2109

PR #1661, opened 2023-05-17
fix(docs): fix open components in a new tab #1641

paghar:
#1641


smolenski-mikolaj:
Actually the issue refers to our docs and internal section links, not the demo application.

PR #1739, opened 2023-06-29
Update calculation of element offsets

### LINKED ISSUE #1738: Scrollspy items not activated correctly when page loads with scroll position not at the top of the page
When a page with scrollspy is loaded with the scroll position somewhere in the middle of the spied element (e.g. when the url has an anchor), the last nav link is always the one that's activated, and the scrollspy doesn't activate the corresponding links as you scroll up and down the page.
However, when the page is loaded with the scroll position at the top of the page, the scrollspy functions properly.

Scrollspy initialization JS:
```
import { ScrollSpy, Collapse, Modal, initTE } from "tw-elements";

initTE({ ScrollSpy, Collapse, Modal });

const profileScrollSpy = new ScrollSpy(document.body, {
        target: "#profile-nav",
        offset: 200,
    });
```

I am using NPM with Vite.

Error occurs on Chrome and Firefox (only browsers I've tested with so far)

AngeloKarugo:
Change the way element top/left offsets are computed, from using document.body scroll attributes (scrollTop and scrollLeft) to get scroll position, to using document.documentElement scroll attributes.
This is because document.body scroll attributes always return zero unless the body has a fixed height and overflow set to auto/scroll [as explained here](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTop).
document.documentElement scroll attributes resolve to scroll attributes of the window [as explained here](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTop#:~:text=When%20scrollTop%20is%20used%20on%20the%20root%20element%20(the%20%3Chtml%3E%20element)%2C%20the%20scrollY%20of%20the%20window%20is%20returned.%20This%20is%20a%20special%20case%20of%20scrollTop).
fixes #1738

### DIFF
--- a/src/js/dom/manipulator.js
+++ b/src/js/dom/manipulator.js
@@ -96,8 +96,8 @@ const Manipulator = {
     const rect = element.getBoundingClientRect();
 
     return {
-      top: rect.top + Math.max(document.body.scrollTop, document.documentElement.scrollTop, window.scrollY),
-      left: rect.left + Math.max(document.body.scrollLeft, document.documentElement.scrollLeft, window.scrollX),
+      top: rect.top + document.body.scrollTop,
+      left: rect.left + document.body.scrollLeft,
     };
   },
 
@@ -150,6 +150,27 @@ const Manipulator = {
 
   hasClass(element, className) {
     return element.classList.contains(className);
+  },
+
+  maxOffset(element) {
+    const rect = element.getBoundingClientRect();
+
+    return {
+      top:
+        rect.top +
+        Math.max(
+          document.body.scrollTop,
+          document.documentElement.scrollTop,
+          window.scrollY
+        ),
+      left:
+        rect.left +
+        Math.max(
+          document.body.scrollLeft,
+          document.documentElement.scrollLeft,
+          window.scrollX
+        ),
+    };
   },
 };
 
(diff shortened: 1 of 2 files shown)


juujisai:
Hi, thank you for your Pull Request. Your updates to Manipulator fixes the scrollspy but I am afraid they may create issues in existing and future components (I've seen `animations` not working properly). I would like to suggest some changes:
- instead of changing the `offset` method from the `Manipulator`, create a new method called `maxOffset` with the changes to calculating top and left position. The `offset` method should stay as it was before
```
  maxOffset(element) {
    const rect = element.getBoundingClientRect();

    return {
      top:
        rect.top +
        Math.max(
          document.body.scrollTop,
          document.documentElement.scrollTop,
          window.scrollY
        ),
      left:
        rect.left +
        Math.max(
          document.body.scrollLeft,
          document.documentElement.scrollLeft,
          window.scrollX
        ),
    };
  },
```
- in `scrollspy.js` changing the `METHOD_OFFSET` constant from `offset` to `maxOffset`
```
const METHOD_OFFSET = "maxOffset";
```

After doing that we would make sure, only the scrollspy will get the updated way of calculating `top and left` values. 

Best Regards!

AngeloKarugo:
@juujisai thank you for the pointers to a more robust solution!
I've made the updates as per you recommendations.

PR #1862, opened 2023-08-12
Fix issue #1852: Hide placeholder text by default and show on focus

abdulhaseeb-com:
This commit addresses the issue where the placeholder text was always visible in the input field. The placeholder text is now hidden by default and will only appear when the input field is focused. The JavaScript functions have been added to control the visibility of the placeholder text based on the input's focus state. This enhances the user experience by providing a cleaner and more intuitive input interaction.
Transition Error Page
https://tailwind-elements.com/angular/

### DIFF
--- a/site/content/angular/index.html
+++ b/site/content/angular/index.html
@@ -17,6 +17,15 @@ menu:
 ---
 
 <!-- Section: Intro -->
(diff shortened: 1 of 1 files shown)


juujisai:
Hi! Thank you for your feature request, but we won't be making changes to the waiting list pages.

PR #1956, opened 2023-09-20
Update quickstart npm docs to warn for monorepo pitfall

frankmayer:
Update quickstart npm docs to warn users of a potential  monorepo pitfall.
In my case I was testing Bun in parallel to my main pnpm/turborepo environment, and tw-elements was not working when packages were installed with bun.
It turned out, bun was only installing it into the root node_modules and not also providing a symlink to the actual app's node_modules. As no error messages were logged, it took some time to find the culprit, so I wanted to spare other users the pain :)

### DIFF
--- a/site/content/docs/standard/getting-started/quick-start/index.html
+++ b/site/content/docs/standard/getting-started/quick-start/index.html
@@ -214,6 +214,19 @@ menu:
   {{< /twsnippet/code >}}{{< /twsnippet/wrapper >}}
 </div>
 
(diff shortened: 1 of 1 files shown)


frankmayer:
@cla-bot check

juujisai:
Hi! Thank you for your PR. I think it will be better if we add a new section with installation guide for bun. 

https://github.com/mdbootstrap/TW-Elements/issues/2064

PR #1962, opened 2023-09-22
Update index.html

heydamianc:
Corrects documentation.


juujisai:
Thank you for your PR. We are reworking integration tutorials so it will be fixed by us

PR #2007, opened 2023-10-17
Fixed chart example squished on mobile view Documentation

njolipatrick:


njolipatrick:
@cla-bot check

juujisai:
Hello, sorry for the late response. Thank you for helping us to make the TW Elements better.

- Charts look ok in mobile but are a little too big in desktop mode. Maybe we could add a breakpoint for the width? `md:w-3/5 w-11/12`
- I would also like to do the same with places where you changed padding: `md:p-6 p-2`

Would you be up to make those changes?

juujisai:
I'll merge this PR to the `fix-squished-charts-examples` brancha and make the updates myself. Thanks again for submitting the PR!

PR #2100, opened 2023-11-08
Change the way ripple gets it's default BS values

juujisai:


juujisai:
https://github.com/mdbootstrap/TW-Elements/issues/2012

PR #2109, opened 2023-11-09
Fix issues with links from api section not showing proper subsections after clicking on a link

juujisai:


juujisai:
https://github.com/mdbootstrap/TW-Elements/issues/1641

PR #2151, opened 2023-11-20
Select: remove name inherit for Input

### LINKED ISSUE #1986: Select - check if input needs to inherit `name` attribute
- most likely was needed for validation
- best case scenario - remove `name` attribute from input 
- check if select collects current value

Rel - https://github.com/mdbootstrap/Tailwind-Elements/issues/1984

Trochonovitz:


juujisai:
@cla-bot check

PR #2221, opened 2023-12-06
remove class from SVG templates so that icons can be resized via wrap…

fero8:
…per claseses

class="w-4 h-4" removed from the icon templates. Hardcoded with & height tailwind classes prevent the possibility to adjust the icon size via the wrapper classes.

### DIFF
--- a/src/js/components/lightbox.js
+++ b/src/js/components/lightbox.js
@@ -35,31 +35,31 @@ const ATTR_LIGHTBOX_CAPTION = `data-te-lightbox-caption`;
 const ATTR_LIGHTBOX_DISABLED = `data-te-lightbox-disabled`;
 const ATTR_STATE_ACTIVE = `data-te-lightbox-active`;
 
-const prevIconTemplate = `<svg xmlns="http://www.w3.org/2000/svg" fill="currentColor" viewBox="0 0 24 24" stroke-width="3" stroke="currentColor" class="w-4 h-4">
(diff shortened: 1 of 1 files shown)


fero8:
@cla-bot check

juujisai:
Hi! Thank you for your Pull Request! 

We cannot remove those styles because the look (size) of those icons would be wrong. Those icons would get their full size and would look too big. I'm afraid we won't be mergin this PR to the project.

There are 2 ways we can customize those arrows styles:

1. via changing the classNames. We can add `[&>svg]:w-2 [&>svg]:h-2` classes to change the size of those arrows. Example:

```
<div
  data-te-lightbox-init
  data-te-class-next-btn="border-none bg-transparent w-full h-[50px] flex justify-center items-center text-[#b3b3b3] transition-colors duration-200 ease-in-out hover:text-white focus:text-white motion-reduce:transition-none outline-none [&>svg]:w-2 [&>svg]:h-2"
...
```

2. by changing the template and updating those classes there

```
<div
  data-te-lightbox-init
  data-te-next-icon-template='<svg xmlns="http://www.w3.org/2000/svg" fill="currentColor" viewBox="0 0 24 24" stroke-width="3" stroke="currentColor" class="w-2 h-2">
    <path stroke-linecap="round" stroke-linejoin="round" d="M4.5 12h15m0 0l-6.75-6.75M19.5 12l-6.75 6.75" />
  </svg>'
  class="flex flex-col space-y-5 lg:flex-row lg:space-x-5 lg:space-y-0">
```

Issue #2294, opened 2024-01-26
Autocomplete: clicking on `No results found` field causes an error


iprzybysz:
[Issue #2293](https://github.com/mdbootstrap/TW-Elements/issues/2293)

PR #2297, opened 2024-01-29
fix: issue #2294 ## Clicking on No results found field causes an error

MONSURU-rgb:
The bug on Form Autocomplete field was resolved by conditionally checking for Null and Undefined values before invoking the getAttribute method.


juujisai:
Thank you for your PR, we will fix this ourselves though

PR #2316, opened 2024-02-21
fix spelling

harlantwood:


juujisai:
Thank you very much for the PR but we will not be merging PRs to this repo

